PR #2, opened 2023-06-02
Refactor and cleaning up initial commit.

chanchiwai-ray:
Refactor and cleaning up initial commit.

- Fix typing issues in `prometheus_hardware_exporter/`
- Fix unit tests after cleaning up  `prometheus_hardware_exporter/`
- Fix functional tests: remove actual metrics tests because it depends on the hardware; only test snap installation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,10 @@ configs = {
     },
 }
 
+with open("requirements.txt", encoding="utf-8") as f:
+    requirements = [req.strip() for req in f.readlines()]
+    configs.update({"install_requires": requirements})
+
 with open("LICENSE", encoding="utf-8") as f:
     configs.update({"license": f.read()})
 
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -9,17 +9,6 @@ TMP_DIR = "/tmp"
 SETUP_TIMEOUT = 5
 
 
-@pytest.fixture()
-def available_metric_names():
-    return [
-        "megaraid_controllers",
-        "megaraid_virtual_drive",
-        "megaraid_virtual_drive_state",
-        "megaraid_virtual_drive_cache_policy",
-        "storcli_command_success",
-    ]
-
-
 @pytest.fixture(scope="session")
 def snap_name():
     return "prometheus-hardware-exporter"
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -7,6 +7,7 @@
     "description": "exports hardware related metrics",
     "use_scm_version": True,
     "setup_requires": ["setuptools_scm", "pyyaml"],
+    "install_requires": ["pyyaml", "pydantic", "prometheus_client"],
```
rgildein, line 10: we should refer to `requirements.txt` to avoid to have two place with same requirements.

rgildein (changes requested):

setup.py
chanchiwai-ray: I am not quite sure about this either. I thought the `install_requires` and `requirement.txt` are for different purposes. https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py

rgildein: I think that in our case we do not take advantage of that, so I do not see point to have same requirement in two places and if you want to change one you will need to change it in two places. For us requirements in requirements.txt are minimal requirements needed for these to work.

setup.py
```diff
@@ -17,6 +17,10 @@
     },
 }
 
+with open("requirements.txt", encoding="utf-8") as f:
```
rgildein, line 20: not-blocking: I'll suggest to use setup.cfg, which brings multiple advantages such reading files (requirements.txt, LICENSE) with simple `file: <path>`, requirements + extra (for litn, unit, func tests or building documentation) requirements could be define in single place.

tests/functional/conftest.py
```diff
@@ -40,7 +29,9 @@ def setup_snap(snap_name):
     if test_snap:
         logging.info("Installing %s snap package...", test_snap)
         assert os.path.isfile(test_snap)
-        assert check_call(f"sudo snap install --dangerous {test_snap}".split()) == 0  # noqa
+        assert (
+            check_call(f"sudo snap install --dangerous --classic {test_snap}".split()) == 0
+        )  # noqa
```
rgildein, line 34: Are these brackets and `# noqa` for black to be happy?

rgildein (approved):
LGTM, one question

setup.py
chanchiwai-ray: I suggest we hold off that because we will lose the ability to use scm versioning. And moving from `setup.py` to `setup.cfg` (both are legacy) may not be worthy.

sudeephb (approved):
LGTM

PR #3, opened 2023-06-06
Add collectors for ipmi metrics.

sudeephb:
The metrics created from the results of the following commands are included
* ipmi-sel
* ipmimonitoring
* ipmi-dcmi


prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +112,261 @@ def fetch(self) -> List[Payload]:
...
+    ipmimonitoring = IpmiMonitoring()
+
+    @property
+    def specifications(self) -> List[Specification]:
+        """Define ipmi sensors metrics specs."""
+        return [
+            Specification(
+                name="ipmi_temperature_celsius",
+                documentation="Temperature measure from temperature sensors",
+                labels=["id", "name", "state", "unit", "event"],
```
Pjack, line 175: What is "event"? Will it contains many different value? 
If it is, I will suggest to remove it to reduce the dimension. 
Similar situation to other metrics. 
What do you think? 
We can also host a small discussion to discuss how to handle "event" if it is still a mandatory information for the metrics.

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +112,261 @@ def fetch(self) -> List[Payload]:
...
+                metric_class=GaugeMetricFamily,
+            ),
```
Pjack, line 177: Another question: 
If we want to draw a diagram on grafana: historical data of temperature of each sensor (1 ID 1 curve line), is it possible? no matter how many state there.

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +112,261 @@ def fetch(self) -> List[Payload]:
...
+    def fetch(self) -> List[Payload]:
+        """Load ipmi sensors data."""
+        sensor_data = self.ipmimonitoring.get_sensor_data()
+
+        if not sensor_data:
+            logger.error("Failed to get ipmi sensor data.")
+            return [Payload(name="ipmimonitoring_command_success", labels=[], value=0.0)]
+
+        payloads = [Payload(name="ipmimonitoring_command_success", labels=[], value=1.0)]
+        for sensor_data_item in sensor_data:
```
Pjack, line 231: It's interesting that we can only identify the items from "Units" type..  
It may be wrong some day. Is there any better way?

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +112,261 @@ def fetch(self) -> List[Payload]:
...
+    ipmi_sel = IpmiSel()
+
+    @property
+    def specifications(self) -> List[Specification]:
+        """Define IPMI SEL metrics specs."""
+        return [
+            Specification(
+                name="ipmi_sel_state",
+                documentation="Event state from IPMI SEL entry.",
+                labels=["id", "date", "time", "name", "type", "event"],
```
Pjack, line 330: ID / date / time / event seems changes every time. ID will be a increment counter. I think we should collect the event in Loki instead of Prometheus.
Therefore, I also purpose to remove these labels to reduce the dimentions.

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
sudeephb: Yes, it is possible since prometheus stores the temperature gauge.

prometheus_hardware_exporter/collector.py
sudeephb: Another way could be identifying by the `type` field of the sensor data item. But, since the format of this data is part of standard defined by IPMI, it's unlikely it will be changed often/without notice. [I took reference from here](https://github.com/prometheus-community/ipmi_exporter/blob/master/collector_ipmi.go#L173) for this implementation.

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +112,261 @@ def fetch(self) -> List[Payload]:
...
+        return payloads
+
+
+class IpmiSensorsCollector(BlockingCollector):
+    """Collector for ipmi sensors data."""
+
+    ipmimonitoring = IpmiMonitoring()
+
+    @property
+    def specifications(self) -> List[Specification]:
```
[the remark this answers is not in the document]
Pjack: According to this document
https://manpages.ubuntu.com/manpages/jammy/en/man8/ipmimonitoring.8.html

There is only 3 possible value of state, so the alert rule be easily monitored by these 3 state. 

ipmi_generic_sensor_value{state="Normal"}
ipmi_generic_sensor_value{state="Critical"}
ipmi_generic_sensor_value{state="Warning"}

https://github.com/sudeephb/charm-prometheus-hardware-exporter/blob/master/src/prometheus_alert_rules/ipmi_sensors.yaml#L149

no concern here , so close this discussion

prometheus_hardware_exporter/__main__.py
```diff
@@ -35,7 +40,10 @@ def main() -> None:
     root_logger.setLevel(logging.getLevelName(config.level))
 
     exporter = Exporter(config.port)
-    exporter.register(MegaRAIDCollector())
+    exporter.register(MegaRAIDCollector(args.config))
```
chanchiwai-ray, line 43: There's no need to add `args.config` if not using it; its optional

prometheus_hardware_exporter/collectors/ipmi_dcmi.py
```diff
@@ -0,0 +1,37 @@
...
+
+logger = getLogger(__name__)
+
+CURRENT_POWER_REGEX = re.compile(r"^Current Power\s*:\s*(?P<value>[0-9.]*)\s*Watts.*")
+
+
+class IpmiDcmi(Command):
+    """Command line tool for ipmi dcmi."""
+
+    prefix = "sudo"
```
chanchiwai-ray, line 17: There's no need to prefix `sudo`. The package is expected to be used as root user. We should probably warn user that they should run as root somehow

jneo8: Discuss:
I have different opinion for this one. We can't expected user always run it as root user. The exporter should be workable for any user which has sudo permission.
The exporter itself should not relate on charm or snap.

tests/unit/test_collector.py
```diff
@@ -41,3 +168,71 @@ def test_01_mega_raid_collector_installed_and_okay(self):
         self.assertEqual(len(list(payloads)), len(available_metrics))
         for payload in payloads:
             self.assertIn(payload.name, available_metrics)
+
+    def test_10_ipmi_dcmi_collector_not_installed(self):
+        """Test ipmi dcmi collector when ipmi-dcmi is not installed."""
+        ipmi_dcmi_collector = IpmiDcmiCollector()
+        payloads = ipmi_dcmi_collector.collect()
```
chanchiwai-ray, line 175: The command should actually be mocked. Because without it, the `collect()` will try to run the command under user's local environment (credit: @jneo8 ) Same goes to other tests.

chanchiwai-ray (changes requested):
Small issue in unit tests, otherwise LGTM

prometheus_hardware_exporter/collectors/ipmi_dcmi.py
chanchiwai-ray: I think it's the same case as `sosreport`. It will failed if the user is not running it as `root` or with `sudo`. This is an adminstration tool and should be only used by administrator.

tests/unit/test_collector.py
jneo8: The external input should always be mocked to avoid the unexpected side effect.

tests/unit/test_collector.py
```diff
@@ -2,10 +2,135 @@
...
 from prometheus_hardware_exporter.collector import (
+    IpmiDcmiCollector,
+    IpmiSelCollector,
+    IpmiSensorsCollector,
     LSISASControllerCollector,
     MegaRAIDCollector,
 )
 
+SAMPLE_IPMI_SEL_ENTRIES = [
+    {
```
chanchiwai-ray, line 13: Wouldn't it be nicer to put it as a file, and load it when used?

chanchiwai-ray (changes requested):

tests/unit/test_collector.py
chanchiwai-ray: it's non-blocking

chanchiwai-ray (approved):
LGTM

Pjack (approved):
LGTM

PR #4, opened 2023-06-07
Add collector for LSI SAS-2/3 controllers

chanchiwai-ray:
Add collector for LSI SAS-2/3 controllers.


Pjack (changes requested):
The code looks good and modulrized. But I want to have a discussion about the metrics naming and format before approve this MR. thanks

rgildein (dismissed):
LGTM

prometheus_hardware_exporter/collectors/sasircu.py
```diff
@@ -0,0 +1,223 @@
...
+import re
+from collections import defaultdict
+from logging import getLogger
+from typing import Any, Dict
+
+from ..utils import Command
+
+logger = getLogger(__name__)
+
+SECTIONS_REGEX = re.compile(
```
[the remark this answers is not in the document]
chanchiwai-ray: Please refer to the unit tests. I've included some examples to test the regex.

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +110,293 @@ def fetch(self) -> List[Payload]:
...
+                value=1.0,
+            ),
+        ]
+
+        for idx, info in all_information:
+            # Add integrated RAID volume metrics
+            if info["ir_volumes"]:
+                payloads.append(
+                    Payload(
+                        name="lsi_sas_2_ir_volumes",
```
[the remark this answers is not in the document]
chanchiwai-ray: The suffix of `_count` / `_total` will suggest that it is a Counter metric, and a counter metric can only increase or reset to zero. This is not the case for us. On a second thought, if this metric is really causing trouble for you, I will suggest that we remove it entirely. Since the same information is contained in `lsi_sas_3_ir_volume_info` metric (i.e. we can build a volume count metric out of it, making use of the uniqueness of the volume_id). Does it work for you?

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +110,293 @@ def fetch(self) -> List[Payload]:
...
+                documentation="Shows the information about the physical device",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="lsi_sas_2_enclosure",  # will append "_info" internally
+                documentation="Show the information about the enclosure",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="sas2ircu_command_success",
```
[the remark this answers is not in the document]
chanchiwai-ray: This is one of the common practice in prometheus to indicate a failed scrape. https://prometheus.io/docs/instrumenting/writing_exporters/#failed-scrapes

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +110,386 @@ def fetch(self) -> List[Payload]:
...
+                labels=["controller_id"],
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="lsi_sas_2_unready_physical_devices",
+                documentation="Number of unready physical devices",
+                labels=["controller_id"],
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
```
Pjack, line 164: According to https://docs.google.com/spreadsheets/d/1KiQ5cS36zfjf8V32RoD1RKx9q01AhgoKvBQRkonQv5o/edit#gid=1149993398
There is a metrics named "lsi_sas_2_physical_devices", but it is not available here.

tests/unit/test_collector.py
```diff
@@ -41,3 +45,182 @@ def test_01_mega_raid_collector_installed_and_okay(self):
...
+                },
+            },
+        }
+
+        lsi_sas_2_collector.sas2ircu.get_adapters.return_value = mock_adapters
+        lsi_sas_2_collector.sas2ircu.get_all_information.return_value = mock_information
+
+        payloads = lsi_sas_2_collector.collect()
+
+        available_metrics = [spec.name for spec in lsi_sas_2_collector.specifications]
```
Pjack, line 133: I'm thinking is it reasonable to fetch the specification from the source code. 
In the most primitive way, we should declare the expected specification here and verify if the output equals to the specification.

But I'm okay with it now, but we need to review the the specification in the source code to confirm it is declared as the detail design specification.

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
chanchiwai-ray: Nice catch! I missed it, Thanks.

prometheus_hardware_exporter/__main__.py
```diff
@@ -3,7 +3,11 @@
 import argparse
 import logging
 
-from .collector import MegaRAIDCollector
+from .collector import (
+    LSISAS2ControllerCollector,
```
esunar, line 7: Not blocking. However it would be useful to consider using class decorators and store all collectors in an array.

prometheus_hardware_exporter/collector.py
```diff
@@ -109,3 +110,397 @@ def fetch(self) -> List[Payload]:
...
+                name="sas2ircu_command_success",
+                documentation="Indicates if the command is successful or not",
+                metric_class=GaugeMetricFamily,
+            ),
+        ]
+
+    def fetch(self) -> List[Payload]:
+        """Load the LSI SAS-2 controllers related information."""
+        adapters = self.sas2ircu.get_adapters()
```
esunar, line 189: this method needs some refactoring. You can extract the comments as functions.

prometheus_hardware_exporter/__main__.py
chanchiwai-ray: Thanks for suggestions. Can you explain more why using a class decorator it useful in this context? As for storing all collectors in an array, I think it might be better to leave it as it is right now since I would expect that in future there will be configuration option to enable / disable individual collector, so it might not be useful to combine them into a single array

prometheus_hardware_exporter/collector.py
chanchiwai-ray: Tried refactoring and reduce repeated code as much as possible.

prometheus_hardware_exporter/__main__.py
esunar: as an example
```python
registry = []

def register(decorated):
    registry.append(decorated)
    return decorated

@register
def foo():
    return 3

@register
def bar():
    return 5
```
 then in main

```python
for collector in registry
    exporter.register(collector)
```
if there is a config option in future you can just edit the register function to consider config values. In any case you will not need to modify another file when you add a new collector.

tests/unit/test_collector.py
```diff
@@ -41,3 +44,182 @@ def test_01_mega_raid_collector_installed_and_okay(self):
         self.assertEqual(len(list(payloads)), len(available_metrics))
         for payload in payloads:
             self.assertIn(payload.name, available_metrics)
+
+    def test_10_lsi_sas_2_collector_not_installed(self):
+        """Test LSI SAS 2 collector when sas2ircu is not installed."""
+        lsi_sas_2_collector = LSISASControllerCollector(2)
+        payloads = lsi_sas_2_collector.collect()
```
jneo8, line 51: The response here should be mock?
I know the there won't be response if the command is not installed. But it's relate on the user's local environment. Right?

jneo8 (changes requested):
Only one question for unit test.

tests/unit/test_collector.py
chanchiwai-ray: Yes

jneo8 (approved):

prometheus_hardware_exporter/__main__.py
chanchiwai-ray: Okay, that make senses. We should probably include this for future PR when config option for enable / disable collectors is added.

sudeephb (approved):
Overall LGTM

Pjack (approved):
LGTM

PR #5, opened 2023-06-14
feat(perccli): Add collector for PowerEdge RAID

jneo8:
- Add collector for PowerEdge RAID


tests/unit/test_collector.py
```diff
@@ -41,3 +46,78 @@ def test_01_mega_raid_collector_installed_and_okay(self):
...
+            # 1 success, 1 fail
+            mock_cli.ctrl_exists.return_value = True
+            mock_cli.ctrl_successes.return_value = {0: False, 1: True}
+            mock_cli.get_controllers.return_value = {"hostname": "abc.com", "count": 1}
+            mock_cli.get_virtual_drives.return_value = {}
+
+            power_edge_collector = PowerEdgeRAIDCollector()
+            payloads = list(power_edge_collector.collect())
+        assert len(payloads) >= 2
+        print(payloads)
```
chanchiwai-ray, line 61: Can you remove this `print`?

prometheus_hardware_exporter/collector.py
```diff
@@ -5,12 +5,134 @@
...
+
+    perccli = PercCLI()
+
+    @property
+    def specifications(self) -> List[Specification]:
+        return [
+            Specification(
+                name="poweredgeraid_controllers",
+                documentation="Total number of controllers",
+                labels=["hostname"],
```
chanchiwai-ray, line 26: I actually drop it in my revamp of MegaRAID collector because there's `instance` label given by prometheus, which is the same as `hostname`

prometheus_hardware_exporter/collector.py
```diff
@@ -5,12 +5,134 @@
...
+            Payload(
+                name="perccli_command_success",
+                labels=[],
+                value=1.0 if success else 0.0,
+            ),
+        )
+        if not success:
+            return payloads
+
+        seccesses = self.perccli.ctrl_successes()
```
chanchiwai-ray, line 75: `successes` ?

prometheus_hardware_exporter/collector.py
```diff
@@ -5,12 +5,134 @@
...
+                name="poweredgeraid_controllers",
+                labels=[controller_payload["hostname"]],
+                value=controller_payload["count"],
+            )
+        )
+        for ctrl_id, vd_payloads in virtual_device_payload.items():
+            payloads.append(
+                Payload(
+                    name="poweredgeraid_virtual_drive",
+                    labels=[str(ctrl_id)],
```
chanchiwai-ray, line 99: nitpick: if `ctrl_id` is used in many places, maybe it's worthy to return a str of `ctrl_id` or `ctrl_id = str(ctrl_id)` at the begining of the loop

prometheus_hardware_exporter/collectors/perccli.py
```diff
@@ -0,0 +1,132 @@
...
+        return result
+
+    def _cmd_status(self, controller: Dict[str, Any]) -> bool:
+        if controller["Command Status"]["Status"] == "Success":
+            return True
+        return False
+
+    def _ctrl_exists(self, controllers: Dict[str, Any]) -> bool:
+        if (
+            controllers["Controllers"][0]["Command Status"]["Status"] == "Failure"
```
chanchiwai-ray, line 36: nitpick: It's not quite readable. Maybe consider creating variables for status and command status?

prometheus_hardware_exporter/collectors/perccli.py
```diff
@@ -0,0 +1,132 @@
...
+        """Return false if command fail or controller not exists."""
+        return self.ctrl_exists()
+
+    def ctrl_exists(self) -> bool:
+        """Check if controller exists."""
+        result = self._get_controllers()
+        if result.error:
+            logger.error(result.error)
+            return False
+        assert isinstance(result.data, dict)
```
chanchiwai-ray, line 56: Nonblocking: It might not be desired to use `assert` or raise error for any collector in the exporter, especially when there are other collectors because other collectors might be still functioning. Also, since the assertion is a type checking and we are already enforcing typing checking with `mypy`, why not consider using `mypy`? Same for other `assert`s

chanchiwai-ray (changes requested):

prometheus_hardware_exporter/collector.py
```diff
@@ -5,12 +5,134 @@
...
+            ),
+            Specification(
+                name="poweredgeraid_virtual_drive",
+                documentation="Number of virtual drives",
+                labels=["controller_id"],
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="perccli_command_success",
+                documentation="Indicates the if command is successful or not",
```
sudeephb, line 37: I think you mean `if the command...`

prometheus_hardware_exporter/collector.py
```diff
@@ -5,12 +5,134 @@
...
+                labels=[],
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="perccli_command_ctrl_success",
+                documentation="Indicates the if command for each controller is successful or not",
```
sudeephb, line 43: Same here, `if the...`

prometheus_hardware_exporter/collectors/perccli.py
```diff
@@ -0,0 +1,132 @@
...
+        assert isinstance(result.data, str)
+        result = get_json_output(result.data)
+        if result.error:
+            return result
+        return result
+
+    def _cmd_status(self, controller: Dict[str, Any]) -> bool:
+        if controller["Command Status"]["Status"] == "Success":
+            return True
+        return False
```
sudeephb, line 32: We can achieve the same with simply
```
def _cmd_status(self, controller: Dict[str, Any]) -> bool:
    return controller["Command Status"]["Status"] == "Success"
```

sudeephb (commented):
Let a few small comments

prometheus_hardware_exporter/collector.py
jneo8: ack

prometheus_hardware_exporter/collector.py
jneo8: typo

prometheus_hardware_exporter/collectors/perccli.py
jneo8: Creating a variable for this kind of check is not that worthy. But I can add an example output json to the doc-string for better understanding.
How do you think?

chanchiwai-ray: Sorry for not being clear, what I mean is, for example `status = controllers["Controllers"][0]["Command Status"]["Status"]`, then `if status == "Failure"`. It take 3 lines to read the `if` condition

prometheus_hardware_exporter/collector.py
jneo8: Updated

tests/unit/test_collector.py
jneo8: Updated

prometheus_hardware_exporter/collectors/perccli.py
jneo8: Change the structure.

prometheus_hardware_exporter/collector.py
jneo8: Updated

prometheus_hardware_exporter/collector.py
```diff
@@ -5,12 +5,139 @@
...
+
+    perccli = PercCLI()
+
+    @property
+    def specifications(self) -> List[Specification]:
+        return [
+            Specification(
+                name="perccli_command_success",
+                documentation="Indicates if the command is successful or not",
+                labels=[],
```
chanchiwai-ray, line 26: With the PR #4, the labels in Specification and Payload can be omitted if it's []. Same for other places

chanchiwai-ray (changes requested):

prometheus_hardware_exporter/collectors/perccli.py
chanchiwai-ray:
```suggestion
            status = controllers["Controllers"][0]["Command Status"]["Status"]
            description = controllers["Controllers"][0]["Command Status"]["Description"]
            return not (status == "Failure" and description == "No Controller found")
```

prometheus_hardware_exporter/collector.py
jneo8: Updated

chanchiwai-ray (approved):
LGTM

prometheus_hardware_exporter/collector.py
```diff
@@ -5,13 +5,134 @@
...
+
+    @property
+    def specifications(self) -> List[Specification]:
+        return [
+            Specification(
+                name="perccli_command_success",
+                documentation="Indicates if the command is successful or not",
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
```
Pjack, line 29: I didn't find this one in the spec
https://docs.google.com/spreadsheets/d/1KiQ5cS36zfjf8V32RoD1RKx9q01AhgoKvBQRkonQv5o/edit#gid=1149993398

prometheus_hardware_exporter/collector.py
```diff
@@ -5,13 +5,134 @@
...
+                name="perccli_command_ctrl_success",
+                documentation="Indicates if the command for each controller is successful or not",
+                labels=["controller_id"],
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="poweredgeraid_controller_exists",
```
Pjack, line 36: 1. 
What's different between poweredgeraid_controllers and poweredgeraid_controller_exists ?

exists is true/false,  which I think it's more likely an alert rule , not a metrics. I prefer to avoid boolean metrics unless it is neccessary. 

2. this metrics is not available in https://docs.google.com/spreadsheets/d/1KiQ5cS36zfjf8V32RoD1RKx9q01AhgoKvBQRkonQv5o/edit#gid=1149993398

prometheus_hardware_exporter/collector.py
```diff
@@ -5,13 +5,134 @@
...
+                name="poweredgeraid_virtual_drive_info",
+                documentation="Indicates the state of virtual drive",
+                labels=[
+                    "controller_id",
+                    "device_group",
+                    "virtual_drive_id",
+                    "state",
+                    "cache_policy",
+                ],
+                metric_class=GaugeMetricFamily,
```
Pjack, line 61: 1. What's the different if we use InfoFamily? 
2. The name on the spec is "poweredgeraid_virtual_drive_state"

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
jneo8: 1. As I understand from [here](https://github.com/prometheus/client_python/blob/master/prometheus_client/metrics_core.py#L359), it always return 1 but it will put the value as part of the labels. I will saw the result is the same because we already put all the items in the label and the value is already 1.
2. I haven't update spec after the change

Pjack: I prefer to align the practice in the team if possible. And I vote InfoMetricFamily, it makes the behavior (value is always 1) implicit.

jneo8: I update with InfoMetricFamily.

chanchiwai-ray (approved):
Overall no big issues. You may want to rebase and solve the conflicts

Pjack (approved):

PR #6, opened 2023-06-14
Revamp MegaRAID collector and update metrics to suit our purpose

chanchiwai-ray:
Revamp the MegaRAID collector using the JSON output of the `storcli` command, this is more robust than regex. Also, added some more metrics to meet our internal needs.


Pjack (approved):
LGTM

esunar (approved):

Issue #7, opened 2023-06-15
Need option to enable / disable collectors


chanchiwai-ray:
It will be much better to be able enable or disable the collectors via the configuration option or as command line option

PR #11, opened 2023-06-20
Allow enabling collectors via configuration option.

chanchiwai-ray:
Add feature to enable collector registration via config option.

Closes: #7

### DIFF
--- a/prometheus_hardware_exporter/__main__.py
+++ b/prometheus_hardware_exporter/__main__.py
@@ -3,19 +3,11 @@
 import argparse
 import logging
 
-from .collector import (
-    IpmiDcmiCollector,
-    IpmiSelCollector,
-    IpmiSensorsCollector,
-    LSISASControllerCollector,
-    MegaRAIDCollector,
-    PowerEdgeRAIDCollector,
-    SsaCLICollector,
-)
+from .collector import COLLECTOR_REGISTRIES
 from .config import DEFAULT_CONFIG, Config
 from .exporter import Exporter
 
-root_logger = logging.getLogger()
+logger = logging.getLogger(__name__)
 
 
 def parse_command_line() -> argparse.Namespace:
@@ -40,17 +32,16 @@ def main() -> None:
     """Start the prometheus-hardware-exporter."""
     args = parse_command_line()
     config = Config.load_config(config_file=args.config or DEFAULT_CONFIG)
+
+    root_logger = logging.getLogger()
     root_logger.setLevel(logging.getLevelName(config.level))
 
     exporter = Exporter(config.port)
-    exporter.register(MegaRAIDCollector())
-    exporter.register(IpmiDcmiCollector())
-    exporter.register(IpmiSensorsCollector())
-    exporter.register(IpmiSelCollector())
-    exporter.register(LSISASControllerCollector(2))
-    exporter.register(LSISASControllerCollector(3))
-    exporter.register(PowerEdgeRAIDCollector())
-    exporter.register(SsaCLICollector())
+    enabled_collectors = set(config.enable_collectors)
+    for name, collector in COLLECTOR_REGISTRIES.items():
+        if name.lower() in enabled_collectors:
+            logger.info("collector %s is enabled", name)
+            exporter.register(collector)
     exporter.run()
 
 
--- a/prometheus_hardware_exporter/collector.py
+++ b/prometheus_hardware_exporter/collector.py
@@ -16,6 +16,8 @@ from .core import BlockingCollector, Payload, Specification
 
 logger = getLogger(__name__)
 
+__all__ = ["COLLECTOR_REGISTRIES"]
+
 
 class PowerEdgeRAIDCollector(BlockingCollector):
     """Collector for PowerEdge RAID controller."""
@@ -832,3 +834,15 @@ class SsaCLICollector(BlockingCollector):
     def process(self, payloads: List[Payload], datastore: Dict[str, Payload]) -> List[Payload]:
         """Process the payload if needed."""
         return payloads
+
+
+COLLECTOR_REGISTRIES = {
+    "mega-raid-collector": MegaRAIDCollector(),
+    "ipmi-dcmi-collector": IpmiDcmiCollector(),
+    "ipmi-sensor-collector": IpmiSensorsCollector(),
+    "ipmi-sel-collector": IpmiSelCollector(),
+    "lsi-sas-2-collector": LSISASControllerCollector(2),
+    "lsi-sas-3-collector": LSISASControllerCollector(3),
+    "poweredge-raid-collector": PowerEdgeRAIDCollector(),
+    "hpe-ssa-collector": SsaCLICollector(),
+}
--- a/prometheus_hardware_exporter/config.py
+++ b/prometheus_hardware_exporter/config.py
@@ -1,4 +1,4 @@
-"""Module for hardware related configuration."""
+"""Module for hardware exporter related configuration."""
 
 import os
 from logging import getLogger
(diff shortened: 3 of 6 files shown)


prometheus_hardware_exporter/__main__.py
```diff
@@ -42,13 +35,12 @@ def main() -> None:
...
-    exporter.register(IpmiDcmiCollector())
-    exporter.register(IpmiSensorsCollector())
-    exporter.register(IpmiSelCollector())
-    exporter.register(LSISASControllerCollector(2))
-    exporter.register(LSISASControllerCollector(3))
-    exporter.register(PowerEdgeRAIDCollector())
+
+    collector_filter = set(config.disable_collector)
+    for name, collector in collector_registries.items():
+        if name.lower() in collector_filter:
```
rgildein, line 41: I think it would be nice to have some log message here. e.g. `logger.info("collector %s is disabled", name)`

prometheus_hardware_exporter/collector.py
```diff
@@ -733,3 +735,14 @@ def fetch(self) -> List[Payload]:
     def process(self, payloads: List[Payload], datastore: Dict[str, Payload]) -> List[Payload]:
         """Process the payload if needed."""
         return payloads
+
+
+collector_registries = {
```
rgildein, line 740: This is constant, so I think it would be better to make it uppercase.

prometheus_hardware_exporter/config.py
```diff
@@ -37,6 +39,28 @@ def validate_level_choice(cls, level: str) -> str:  # noqa: N805 pylint: disable
             raise ValueError(msg)
         return level
 
+    @validator("disable_collector")
+    def validate_disable_collector_choice(  # pylint: disable=E0213
+        cls, disable_collector: List[str]  # noqa: N805
```
rgildein, line 44: Why using cls here? This is not class method or it is?

prometheus_hardware_exporter/config.py
```diff
@@ -37,6 +39,28 @@ def validate_level_choice(cls, level: str) -> str:  # noqa: N805 pylint: disable
             raise ValueError(msg)
         return level
 
+    @validator("disable_collector")
+    def validate_disable_collector_choice(  # pylint: disable=E0213
```
rgildein, line 43: Same here why we are using cls and providing two ingore here?

prometheus_hardware_exporter/__main__.py
jneo8: +1

prometheus_hardware_exporter/collector.py
jneo8: +1

jneo8 (commented):
Same as Robert.

prometheus_hardware_exporter/config.py
chanchiwai-ray: Please refer to [pydantic's documentation](https://docs.pydantic.dev/latest/usage/validators/)

prometheus_hardware_exporter/config.py
chanchiwai-ray: Please refer to [pydantic's documentation](https://docs.pydantic.dev/latest/usage/validators/)

prometheus_hardware_exporter/collector.py
chanchiwai-ray: DONE

prometheus_hardware_exporter/__main__.py
chanchiwai-ray: DONE

Pjack (dismissed):

prometheus_hardware_exporter/__main__.py
rgildein: This is not good practice to use root logger to making logs, we should use root_logger only for configuration. At least I think that's good practice. What do you think?

chanchiwai-ray: I am not sure if this is what you mean: are you suggesting that 

```
logger = getLogger(__main__)
root_logger = getLogger() # keeping this for setLevel
...
for name, collector in collector_registries.items():
   if name.lower() in collector_filter:
       logger.info(....)
```
?

prometheus_hardware_exporter/config.py
rgildein: I see, but still I will rather take a look how we can fix it, solution for flake8 is in [#568](https://github.com/pydantic/pydantic/issues/568) issue, or define ignore for whole file instead of doing it in every validate function.
What do you think?

prometheus_hardware_exporter/__main__.py
rgildein: Initially that was my idea, but after checking it we should do.
```python
logger = getLogger(__main__)
... 
for name, collector in collector_registries.items():
   if name.lower() in collector_filter:
       logger.info(....)

def main():
...
    root_logger = logging.getLogger()
    root_logger.setLevel(...)
```

prometheus_hardware_exporter/config.py
chanchiwai-ray: Thanks for the information, I did not go deeper to find out the issue. Applied suggestion.

prometheus_hardware_exporter/__main__.py
chanchiwai-ray: I've updated this part.

jneo8 (dismissed):

chanchiwai-ray:
There is currently some issue with python plugin + classic confinement during `override-build`, I will skip this failure at the moment since we don't have a plan to use snap right now.

prometheus_hardware_exporter/collector.py
```diff
@@ -832,3 +834,14 @@ def fetch(self) -> List[Payload]:
     def process(self, payloads: List[Payload], datastore: Dict[str, Payload]) -> List[Payload]:
         """Process the payload if needed."""
         return payloads
+
+
+COLLECTOR_REGISTRIES = {
```
esunar, line 839: Not blocker but it would be nice to auto register with annotations
`@collector("mega-raid-collector")`

chanchiwai-ray: It's not necessary that one class represent one collector, e.g. we have 

```
    "lsi-sas-2-collector": LSISASControllerCollector(2),
    "lsi-sas-3-collector": LSISASControllerCollector(3),
```

chanchiwai-ray: well, it might work if we do this:

```
COLLECTOR_REGISTRIES= {}

def register(name, *args, **kwargs):
    def decorator(cls):
        COLLECTOR_REGISTRIES[name] = cls(*args, **kwargs)
        return register                                                                                                                     
    return decorator

```

chanchiwai-ray: I've tried it, and should be working, but the type check effectively make it even more complicated. So if it's not a blocker, then I will try to avoid that.
```
COLLECTOR_REGISTRIES= {}

def register(name, *args, **kwargs):
    def decorator(cls):
        COLLECTOR_REGISTRIES[name] = cls(*args, **kwargs) # type checking in here has no simple solution, because each class has different arguments, and require different types.
        return cls                                                                                                                          
    return decorator

```

prometheus_hardware_exporter/config.py
```diff
@@ -11,14 +12,18 @@
 DEFAULT_CONFIG = os.path.join(os.environ.get("SNAP_DATA", "./"), "config.yaml")
 
 
+# pylint: disable=E0213
+
+
 class Config(BaseModel):
     """Juju backup all configuration."""
```
sudeephb, line 19: This docstring needs to be updated.

chanchiwai-ray: DONE

Pjack (approved):

sudeephb (approved):

prometheus_hardware_exporter/__main__.py
rgildein: Thanks :+1:

PR #15, opened 2023-06-28
Improve function test

chanchiwai-ray:
- Add explicit command line option to enable collectors
- Convert function test for snap to functional test for package.


sudeephb (approved):

prometheus_hardware_exporter/__main__.py
```diff
@@ -22,28 +23,84 @@ def parse_command_line() -> argparse.Namespace:
...
+    parser.add_argument("-l", "--level", help="Set logging level.", default="INFO", type=str)
     parser.add_argument("-c", "--config", help="Set configuration file.", default="", type=str)
+    parser.add_argument(
+        "-p",
+        "--port",
+        help="Address on which to expose metrics and web interface.",
+        default=10000,
+        type=int,
+    )
+    parser.add_argument(
```
jneo8, line 35: **suggestion:** 

Why don't just use `nargs="+"` with one argument here? If all the collector is disable by default.

jneo8 (approved):
Just one non-blocking comment.

prometheus_hardware_exporter/__main__.py
chanchiwai-ray: We could, but It will be nicer if we can also provide help string for each collector. And this pattern is also the same for `prometheus-node-exporter`

PR #16, opened 2023-06-29
Add Redfish collector

sudeephb:
This uses [redfish-utilities](https://pypi.org/project/redfish-utilities/). ([Github repo](https://github.com/DMTF/Redfish-Tacklebox)). We will need to incorporate it in the install strategy, hence it's not in `requirements.txt`.

### DIFF
--- a/prometheus_hardware_exporter/__main__.py
+++ b/prometheus_hardware_exporter/__main__.py
@@ -2,9 +2,17 @@
 
 import argparse
 import logging
-from typing import List
 
-from .collector import COLLECTOR_REGISTRIES, RedfishCollector
+from .collector import (
+    IpmiDcmiCollector,
+    IpmiSelCollector,
+    IpmiSensorsCollector,
+    LSISASControllerCollector,
+    MegaRAIDCollector,
+    PowerEdgeRAIDCollector,
+    RedfishCollector,
+    SsaCLICollector,
+)
 from .config import DEFAULT_CONFIG, Config
 from .exporter import Exporter
 
@@ -31,6 +39,24 @@ def parse_command_line() -> argparse.Namespace:
         help="Address on which to expose metrics and web interface.",
         default=10000,
         type=int,
+    )
+    parser.add_argument(
+        "--redfish_host",
+        help="Hostname for redfish collector",
+        default="127.0.0.1",
+        type=str,
+    )
+    parser.add_argument(
+        "--redfish_username",
+        help="BMC username for redfish collector",
+        default="",
+        type=str,
+    )
+    parser.add_argument(
+        "--redfish_password",
+        help="BMC password for redfish collector",
+        default="",
+        type=str,
     )
     parser.add_argument(
         "--collector.hpe_ssa",
@@ -77,14 +103,25 @@ def parse_command_line() -> argparse.Namespace:
     return args
 
 
-def main(port: int, level: str, enable_collectors: List[str], daemon: bool = False) -> None:
+def main(config: Config, daemon: bool = False) -> None:
     """Start the prometheus-hardware-exporter."""
     root_logger = logging.getLogger()
-    root_logger.setLevel(logging.getLevelName(level))
+    root_logger.setLevel(logging.getLevelName(config.level))
 
-    exporter = Exporter(port)
-    enable_collectors_set = set(enable_collectors)
-    for name, collector in COLLECTOR_REGISTRIES.items():
+    exporter = Exporter(config.port)
+    enable_collectors_set = set(config.enable_collectors)
+    collector_registries = {
+        "collector.hpe_ssa": SsaCLICollector(config),
+        "collector.ipmi_dcmi": IpmiDcmiCollector(config),
+        "collector.ipmi_sel": IpmiSelCollector(config),
+        "collector.ipmi_sensor": IpmiSensorsCollector(config),
+        "collector.lsi_sas_2": LSISASControllerCollector(version=2, config=config),
+        "collector.lsi_sas_3": LSISASControllerCollector(version=3, config=config),
+        "collector.mega_raid": MegaRAIDCollector(config),
+        "collector.poweredge_raid": PowerEdgeRAIDCollector(config),
+        "collector.redfish": RedfishCollector(config),
+    }
+    for name, collector in collector_registries.items():
         if name.lower() in enable_collectors_set:
             logger.info("%s is enabled", name)
             exporter.register(collector)
@@ -94,13 +131,13 @@ def main(port: int, level: str, enable_collectors: List[str], daemon: bool = Fal
 if __name__ == "__main__":  # pragma: no cover
     ns = parse_command_line()
     if ns.config:
-        config = Config.load_config(config_file=ns.config or DEFAULT_CONFIG)
+        exporter_config = Config.load_config(config_file=ns.config or DEFAULT_CONFIG)
     else:
         collectors = []
         for args_name, enable in ns.__dict__.items():
             if args_name.startswith("collector") and enable:
                 collectors.append(args_name)
-        config = Config(port=ns.port, level=ns.level, enable_collectors=collectors)
+        exporter_config = Config(port=ns.port, level=ns.level, enable_collectors=collectors)
 
     # Start the exporter
-    main(config.port, config.level, config.enable_collectors)
+    main(exporter_config)
--- a/prometheus_hardware_exporter/__main__.py
+++ b/prometheus_hardware_exporter/__main__.py
@@ -41,19 +41,19 @@ def parse_command_line() -> argparse.Namespace:
         type=int,
     )
     parser.add_argument(
-        "--redfish_host",
+        "--redfish-host",
         help="Hostname for redfish collector",
         default="127.0.0.1",
         type=str,
     )
     parser.add_argument(
-        "--redfish_username",
+        "--redfish-username",
         help="BMC username for redfish collector",
         default="",
         type=str,
     )
     parser.add_argument(
-        "--redfish_password",
+        "--redfish-password",
         help="BMC password for redfish collector",
         default="",
         type=str,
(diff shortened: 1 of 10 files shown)


prometheus_hardware_exporter/collector.py
```diff
@@ -845,4 +916,5 @@ def process(self, payloads: List[Payload], datastore: Dict[str, Payload]) -> Lis
     "lsi-sas-3-collector": LSISASControllerCollector(3),
     "poweredge-raid-collector": PowerEdgeRAIDCollector(),
     "hpe-ssa-collector": SsaCLICollector(),
+    "redfish-collector": RedfishCollector("", "", ""),
```
jneo8, line 919: **issue (non-blocking):** 

I guess here somehow bring out an issue: what if the collector need some config to initialize?

I will suggest to refactor this part in later pr.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get_service_status(self) -> bool:
+        """Return whether redfish service is available or not."""
+        result = self()
+        if result.error:
+            logger.error(result.error)
+            return False
+
+        rf_services = result.data.strip()
+        return "No Redfish services discovered" not in rf_services
```
jneo8, line 66: **suggestion (non-blocking):** 

Can you confirm this string is always working in different situation?
I expect we can get a more contain answer to say it's not available.

jneo8 (changes requested):
Need more information.

prometheus_hardware_exporter/__main__.py
```diff
@@ -41,6 +41,11 @@ def main() -> None:
     for name, collector in COLLECTOR_REGISTRIES.items():
         if name.lower() in enabled_collectors:
             logger.info("collector %s is enabled", name)
+            if name.lower() == "redfish-collector":
+                exporter.register(
+                    RedfishCollector(config.bmc_host, config.bmc_username, config.bmc_password)
+                )
+                continue
```
Pjack, line 48: I prefer to use if/else to show the logic

Pjack: Will collector read the config file internal? If yes, then we don't need to parse the username/password here. 

Another idea is `config` will be a parameter for all collecter.
if config == None => use default value
if config != None => collector will change the behavior.

Pjack (changes requested):

chanchiwai-ray:
Can I ask why we don't use the [native python redfish libarary](https://github.com/DMTF/python-redfish-library) ?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -0,0 +1,66 @@
...
+from ..utils import Command
+
+logger = getLogger(__name__)
+
+
+class RedfishSensors(Command):
+    """Command line tool for getting sensor data using redfish."""
+
+    prefix = ""
+    command = "rf_sensor_list.py"
```
jneo8, line 15: **issue:** 

Where is this file?

jneo8 (changes requested):
Need information

prometheus_hardware_exporter/collectors/redfish.py
chanchiwai-ray: it's in https://github.com/DMTF/Redfish-Tacklebox/tree/main/scripts

sudeephb:
> Can I ask why we don't use the [native python redfish libarary](https://github.com/DMTF/python-redfish-library) ?

We could, but we would be re-doing much of what `redfish-utilities` is doing using the redfish library. 
Also, as a small plus, we can use the `redfish-utilities` as different commands directly, which is consistent with other collectors.

prometheus_hardware_exporter/collectors/redfish.py
sudeephb: Additional note: When we install the `redfish-utilities` library, we can use this as a command directly (as it's in `/usr/local/bin`).

prometheus_hardware_exporter/__main__.py
Pjack: I guess you misunderstand what I mean based on the latest source code. 
I didn't mean to unzip the collector table. I just say not to use `continue`
```python
    for name, collector in COLLECTOR_REGISTRIES.items():
        if name.lower() in enable_collectors_set:
            logger.info("%s is enabled", name)
            if name.lower() == "collector.redfish":
                  exporter.register(
                    RedfishCollector(config.bmc_host, config.bmc_username, config.bmc_password)
                )
            else:
                  exporter.register(collector)
```

prometheus_hardware_exporter/__main__.py
```diff
@@ -77,17 +104,40 @@ def parse_command_line() -> argparse.Namespace:
...
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.getLevelName(level))
 
     exporter = Exporter(port)
     enable_collectors_set = set(enable_collectors)
-    for name, collector in COLLECTOR_REGISTRIES.items():
-        if name.lower() in enable_collectors_set:
-            logger.info("%s is enabled", name)
-            exporter.register(collector)
+    if "collector.hpe_ssa" in enable_collectors_set:
```
Pjack, line 123: These `if` are awful. I guess you misunderstand what I mean in 
https://github.com/canonical/prometheus-hardware-exporter/pull/16#discussion_r1246361637

prometheus_hardware_exporter/collectors/redfish.py
Pjack: Could we put some note here? 
Furthermore, this is a potential risk and we must test it in the functional test. not only unit test.

prometheus_hardware_exporter/__main__.py
```diff
@@ -77,17 +104,40 @@ def parse_command_line() -> argparse.Namespace:
     return args
 
 
-def main(port: int, level: str, enable_collectors: List[str], daemon: bool = False) -> None:
+# pylint: disable=R0913
+def main(
+    port: int,
+    level: str,
+    enable_collectors: List[str],
+    redfish_host: str,
```
Pjack, line 112: redfish_xxx is only used by redfish. 
Is it possible that we can encapsulated the rest of parameters as a config dictionary and pass into collector?

prometheus_hardware_exporter/__main__.py
sudeephb: Oh I got it. I thought you wanted to make the `registration` consistent for all collectors. Anyway, in the next commit(after lint fix), I am using config directly in the collector, so this issue will go away.

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
sudeephb: This is now removed.

prometheus_hardware_exporter/collectors/redfish.py
sudeephb: I added a note with the link in docstring of this function

prometheus_hardware_exporter/__main__.py
```diff
@@ -94,13 +131,13 @@ def main(port: int, level: str, enable_collectors: List[str], daemon: bool = Fal
...
     if ns.config:
-        config = Config.load_config(config_file=ns.config or DEFAULT_CONFIG)
+        exporter_config = Config.load_config(config_file=ns.config or DEFAULT_CONFIG)
     else:
         collectors = []
         for args_name, enable in ns.__dict__.items():
             if args_name.startswith("collector") and enable:
                 collectors.append(args_name)
-        config = Config(port=ns.port, level=ns.level, enable_collectors=collectors)
+        exporter_config = Config(port=ns.port, level=ns.level, enable_collectors=collectors)
```
chanchiwai-ray, line 140: The `Config` is not initialized with redfish related arguments.

chanchiwai-ray (changes requested):

prometheus_hardware_exporter/__main__.py
```diff
@@ -32,6 +40,24 @@ def parse_command_line() -> argparse.Namespace:
         default=10000,
         type=int,
     )
+    parser.add_argument(
+        "--redfish_host",
```
jneo8, line 44: **nitpick:** 

`redfish-host`?

prometheus_hardware_exporter/__main__.py
```diff
@@ -94,13 +136,20 @@ def main(port: int, level: str, enable_collectors: List[str], daemon: bool = Fal
 if __name__ == "__main__":  # pragma: no cover
     ns = parse_command_line()
     if ns.config:
-        config = Config.load_config(config_file=ns.config or DEFAULT_CONFIG)
+        exporter_config = Config.load_config(config_file=ns.config or DEFAULT_CONFIG)
```
jneo8, line 139: **nitpick:** 

`config` instead of `exporter_config` make more clean here, right?

jneo8 (changes requested):

prometheus_hardware_exporter/__main__.py
chanchiwai-ray: `config` is in `__main__` scope. mypy will complain about it, because `main()` function also use `config`

chanchiwai-ray (approved):

prometheus_hardware_exporter/__main__.py
sudeephb: It will result in a lint error in the `main()` function because we're re-defining `config` from the global scope in that function.

prometheus_hardware_exporter/__main__.py
sudeephb: I changed it to `redfish-host` (and also similar for username and password)

Pjack (approved):

jneo8 (approved):

PR #18, opened 2023-06-30
test(collector): make sure collector is register right in cli wth config values

jneo8:


sudeephb (commented):
Overall LGTM, but maybe it's more accurate to remove the `/un-register` from PR description. In the exporter, we don't unregister any collector, we just restart the exporter with the different set of collectors, right?

sudeephb (approved):

tests/unit/test_cli.py
```diff
@@ -21,3 +27,56 @@ def test_cli_main(self, mock_exporter):
...
+
+        for collector in enable_collectors:
+            mock_exporter.return_value.register.assert_any_call(
+                collector_registries.get(collector)
+            )
+
+        for name, collector in collector_registries.items():
+            if name.lower() not in enable_collectors:
+                with pytest.raises(AssertionError):
+                    mock_exporter.return_value.register.assert_called_with(collector)
```
rgildein, line 61: I think that you should mock `get_collector_registries` only and not run it. Maybe you can provide list of collectors in constant and use it also in `test_get_collector_registries`.
```suggestion
    def test_cli_main_enable_collector(self, mock_exporter, enable_collectors):
        config = Config(port=10000, level="INFO", enable_collectors=enable_collectors)
        collector_registries = {
            "collector.hpe_ssa": "collector.hpe_ssa",
            "collector.ipmi_dcmi": "collector.ipmi_dcmi",
            "collector.ipmi_sel": "collector.ipmi_sel",
            "collector.ipmi_sensor": "collector.ipmi_sensor",
            "collector.lsi_sas_2": "collector.lsi_sas_2",
            "collector.lsi_sas_3": "collector.lsi_sas_3",
            "collector.mega_raid": "collector.mega_raid",
            "collector.poweredge_raid": "collector.poweredge_raid",
            "collector.redfish": "collector.redfish",
        }
        
        with patch.object(__main__, "get_collector_registries", return_value=collector_registries):
            main(config)

        mock_exporter.register.assert_has_calls(
            [call(collector) for collector in enable_collectors], any_order=True,
        )
```

rgildein (changes requested):

tests/unit/test_cli.py
jneo8: There is not side effect on `get_collector_registries`, all the collector initialize is clean. I don't think it's that necessary to avoid the call.

tests/unit/test_cli.py
```diff
@@ -21,3 +27,56 @@ def test_cli_main(self, mock_exporter):
...
+
+def test_get_collector_registries():
+    """Assert all the keys."""
+    config = Config(port=10000, level="INFO", enable_collectors=[])
+
+    # Every time someone add/remove collector should change this.
+    collector_registries = get_collector_registries(config)
+    assert len(collector_registries) == 9
+
+    assert list(collector_registries.keys()) == [
```
Pjack, line 72: 1. If we write the unit test more flexibile. We only need to check with `set`. We don't need the function return in a specific order (`list`). After we check the `set` is matched, we don't need to check the length. it's duplicated.

2. We can check if the value is not None.

tests/unit/test_cli.py
Pjack: I'm okay with James' solution. `get_collector_registries` is not the key point here and the correctness will be verified in the next unit test function.

Pjack (changes requested):

tests/unit/test_cli.py
Pjack: Robert's solution is good. But I'm okay with James' solution too. `get_collector_registries` is not the key point here and the correctness will be verified by next unit test function.

Pjack (approved):
LGTM

tests/unit/test_cli.py
rgildein: In general you should mocked any other functions to test single function with single test, but in here you are right it's not necessary. That's why I put only comment and not request changes.

rgildein (commented):
I'm changing it to "comment", because I believed that tests could be simplified. However that is not reason to blocked this.

rgildein (approved):
Changing to "Approve", since submitting "Comment" review will not change the "Request changes".

PR #27, opened 2023-07-12
fix(redfish): Native python redfish

jneo8:
- Use Native python lib of redfish instead of subprocess to call redfish python scripts

The major reason we do this refactor is because `charmcraft` can't package the [scripts file](https://github.com/DMTF/Redfish-Tacklebox/blob/87c231ab16e0bb699460339d18c98b3ac960fde0/setup.py#L29) into charm's bin path. So instead of using the python script, we use redfish & redfish_utilities lib directly.


prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -1,70 +1,65 @@
...
+    def _get_sensor_data(self, host: str, username: str, password: str) -> Optional[List[Any]]:
+        """Return sensor if sensor exists else None."""
+        try:
+            redfish_obj = redfish.redfish_client(
+                base_url=host, username=username, password=password
+            )
+            redfish_obj.login(auth="session")
+            sensors = redfish_utilities.get_sensors(redfish_obj)
+            return sensors
+        except Exception as err:  # pylint: disable=W0718
```
rgildein, line 53: Can we be more specific here and catch error from redfish only.

tests/unit/test_redfish.py
```diff
@@ -1,59 +1,151 @@
...
     """Test the RedfishSensors class."""
 
-    @patch.object(Command, "__call__")
-    def test_00_get_sensor_data_success(self, mock_call):
-        with open(SAMPLE_RF_SENSORS_OUTPUT, "r") as content:
-            mock_call.return_value = Result(content.read(), None)
-            rf_sensors = RedfishSensors()
-            sensor_data = rf_sensors.get_sensor_data("username", "password", "localhost")
-            self.assertEqual(sensor_data, SAMPLE_RF_SENSOR_DATA)
+    def setUp(self):
```
rgildein, line 10: I think that this is not needed or it is?

rgildein (changes requested):

prometheus_hardware_exporter/collectors/redfish.py
jneo8: For this one no, The original source code doesn't raise any specific error.
https://github.com/DMTF/Redfish-Tacklebox/blob/87c231ab16e0bb699460339d18c98b3ac960fde0/redfish_utilities/sensors.py#L15

tests/unit/test_redfish.py
```diff
@@ -1,59 +1,153 @@
...
+            }
+        ],
+    )
+    def test_00_get_sensor_data_success(self, mock_get_sensor_data):
+        helper = RedfishHelper()
+        data = helper.get_sensor_data("", "", "")
+        self.assertEqual(
+            data,
+            {
+                "1": [
```
sudeephb, line 81: Can you add a test case for when there are more than one Chassis?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -1,70 +1,65 @@
...
+            redfish_obj = redfish.redfish_client(
+                base_url=host, username=username, password=password
+            )
+            redfish_obj.login(auth="session")
+            sensors = redfish_utilities.get_sensors(redfish_obj)
+            return sensors
+        except InvalidCredentialsError as err:
+            logger.exception(err)
+        except SessionCreationError as err:
+            logger.exception(err)
```
rgildein, line 54: not-blocking: If we do not add any specific logic to exception why not catch both in one step. I mean:
```suggestion
        except (InvalidCredentialsError, SessionCreationError) as err:
            logger.exception(err)
```

tests/unit/test_collector.py
```diff
@@ -531,20 +531,30 @@ def test_105_perccli_physical_device_command_success(self):
...
 
         self.assertEqual(len(list(payloads)), 2)
 
     def test_201_redfish_installed_and_okay(self):
         """Test redfish collector when redfish-utilitites is installed."""
         redfish_collector = RedfishCollector(Mock())
-        redfish_collector.redfish_sensors = Mock()
-        mock_sensor_data = SAMPLE_RF_SENSOR_DATA
-        redfish_collector.redfish_sensors.get_sensor_data.return_value = mock_sensor_data
+        redfish_collector.redfish_helper = Mock()
```
rgildein, line 544: not-blocking: `redfish_collector.redfish_helper = Mock(spec=RedfishHelper)`

rgildein (approved):
Two not blocking comment, but it LGTM.

sudeephb (approved):

sudeephb (approved):

PR #28, opened 2023-07-12
Remove duplicates and older logs from ipmi sel

sudeephb:


prometheus_hardware_exporter/collector.py
```diff
@@ -547,14 +548,17 @@ def fetch(self) -> List[Payload]:
                     "Unknown ipmi SEL state: %s. Treating it as Nominal.", sel_entry["State"]
                 )
                 sel_state_value = sel_states_dict["NOMINAL"]
+            sel_entries_tuples.add((sel_entry["Name"], sel_entry["Type"], sel_state_value))
+
+        for sel_item in sel_entries_tuples:
```
rgildein, line 553: not-blocking: suggestion
```suggestion
        for sel_name, sel_type, sel_state_value in sel_entries_tuples:
```

tests/unit/requirements.txt
```diff
@@ -1 +1,2 @@
 pytest
+freezegun
```
rgildein, line 2: Is this really required? Could we use mock instead?

Pjack: I am okay with this. It makes the unit test easy to understand.

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
```diff
@@ -547,14 +548,17 @@ def fetch(self) -> List[Payload]:
                     "Unknown ipmi SEL state: %s. Treating it as Nominal.", sel_entry["State"]
                 )
                 sel_state_value = sel_states_dict["NOMINAL"]
+            sel_entries_tuples.add((sel_entry["Name"], sel_entry["Type"], sel_state_value))
```
jneo8, line 551: **question (blocking):** 

If I have two records which have the same name and type, but the state is different: Nominal and Warning.

I will get two metrics? I will suggest to return one record with the urgent state. For this case should be only one Warning record. Multiple records with same name but different state will confuse people.

jneo8 (changes requested):

prometheus_hardware_exporter/collector.py
jneo8: Status will be `Nominal` in most of time. And if the status is `Warning` or `Critical`, user wouldn't care the `Nominal` status. Don't use exporter as logging system. You should provide **single true of source**, which is the most urgent status to user.

jneo8: For lost information. Exporter shouldn't try to provide the information which logging should provide.
And don't provide the duplicate metrics in the exporter. This always confuse people.

tests/unit/requirements.txt
sudeephb: Mocking `datetime.datetime.now()` directly is not possible because it's implemented in C. We could have a workaround, like creating a wrapper object around the datetime object and mocking it. But I think using this tool is simpler.

Pjack (approved):

prometheus_hardware_exporter/collector.py
```diff
@@ -547,13 +548,16 @@ def fetch(self) -> List[Payload]:
                     "Unknown ipmi SEL state: %s. Treating it as Nominal.", sel_entry["State"]
                 )
                 sel_state_value = sel_states_dict["NOMINAL"]
+
+            key = (sel_entry["Name"], sel_entry["Type"])
+            if key not in sel_entries_dict or sel_entries_dict[key] < sel_state_value:
```
jneo8, line 553: **nitpick:** 

I think we miss some tests to verify the merging logic.

jneo8 (changes requested):

prometheus_hardware_exporter/collector.py
sudeephb: All of the merging conditions are tested since the [sample data](https://github.com/sudeephb/prometheus-hardware-exporter/blob/main/tests/unit/test_resources/ipmi/ipmi_sample_data.py#L1) contains all of the possible conditions(repeated data,  repeated data with only the `state` difference).

rgildein (approved):
LGTM

prometheus_hardware_exporter/collector.py
jneo8: How do I know that the value will be `Warning` instead of `Nominal` if there are two records?
The simple way is we check the return payload's details.

jneo8 (approved):

PR #29, opened 2023-07-14
fix(ipmi-sel): ipmi-sel error handling

### LINKED ISSUE #30: ipmi_sel should not return ipmi_sel_command_success 0.0 if there is no new ipmi_sel records
After #28, there is a time period filter to return the ipmi_sel records. Currently the exporter will throw out `ipmi_sel_command_success = 0.0` if there is no new records. But this doesn't mean the command fail status.

We should use another kind of statement to judge if the command is fail or not.

jneo8:
- The PostInit record which doesn't have expected datatime format output
- Fix #30, Return None instead of empty list if command fail.

### DIFF
--- a/prometheus_hardware_exporter/collector.py
+++ b/prometheus_hardware_exporter/collector.py
@@ -532,7 +532,7 @@ class IpmiSelCollector(BlockingCollector):
         sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
 
         if sel_entries is None:
-            logger.warning("No recent ipmi sel entries to collect.")
+            logger.warning("Fail to get ipmi sel entries.")
             return [Payload(name="ipmi_sel_command_success", value=0.0)]
 
         sel_states_dict = {"NOMINAL": 0, "WARNING": 1, "CRITICAL": 2}
--- a/prometheus_hardware_exporter/collector.py
+++ b/prometheus_hardware_exporter/collector.py
@@ -531,7 +531,7 @@ class IpmiSelCollector(BlockingCollector):
         """Load ipmi sel entries."""
         sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
 
-        if not sel_entries:
+        if sel_entries is None:
             logger.warning("No recent ipmi sel entries to collect.")
             return [Payload(name="ipmi_sel_command_success", value=0.0)]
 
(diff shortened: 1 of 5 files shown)


prometheus_hardware_exporter/collector.py
```diff
@@ -531,7 +531,7 @@ def fetch(self) -> List[Payload]:
         """Load ipmi sel entries."""
         sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
 
-        if not sel_entries:
+        if sel_entries is None:
             logger.warning("No recent ipmi sel entries to collect.")
```
sudeephb, line 535: Since this happens only in case of error now, I think we can change the log message
```suggestion
            logger.warning("Failed to get ipmi sel entries.")
```

sudeephb (changes requested):
Thank you for the patch. I have put a small change request in the log message, otherwise LGTM.

sudeephb (approved):

prometheus_hardware_exporter/collector.py
```diff
@@ -531,8 +531,8 @@ def fetch(self) -> List[Payload]:
         """Load ipmi sel entries."""
         sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
 
-        if not sel_entries:
-            logger.warning("No recent ipmi sel entries to collect.")
+        if sel_entries is None:
```
Pjack, line 534: Does None and [] has different meaning here? So that we have to change the syntax here. 
1. There is no description in get_sel_entries function to describe the return value.
2. I prefer to interpret [] and None as the same thing. It's not easy to distinguish them for the following maintainer and make mistake.

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
jneo8: After we have the time range filter, there are multiple possible situations.

- List with sen_entry.
- Empty list.(Fail on command or no records in time range)

Empty list here means *no records in time range*.
So we miss a way of return to represent *Fail on command*.

Use None and empty maybe the easiest way for this purpose. Also because  we have a mypy linter which make us hard to return `Union` type. We need to do more check on the type.

So the solution can be:

- `Union` return `Union[Exception, List]`
- Add one more return value which represent if the cmd success.

Pjack (approved):

PR #34, opened 2023-08-07
Add TTL cache to redfish discover method

### LINKED ISSUE #33: Cache for redfish collector
The current query_timeout for grafana-agent is 10 seconds and sometimes the redfish takes more the 10 seconds to response.

I will suggest to add the cache for it(maybe also other collectors), otherwise we may lose metrics.

dashmage:
The `redfish.discover_ssdp()` method takes a very long time to query the redfish services. This causes the scrape timeout to be exceeded for Prometheus. In order to solve this, a TTL cache has been added. A logging message has also been added for debugging.

Additionally, the `redfish.redfish_client` call to create a redfish object has been provided with the `timeout` and `max_retry` arguments in order to not have the query take too long in case of incorrect redfish credentials.

Fixes #33, canonical/hardware-observer-operator#55

### DIFF
--- a/prometheus_hardware_exporter/__main__.py
+++ b/prometheus_hardware_exporter/__main__.py
@@ -16,7 +16,13 @@ from .collector import (
     RedfishCollector,
     SsaCLICollector,
 )
-from .config import DEFAULT_CONFIG, Config
+from .config import (
+    DEFAULT_CONFIG,
+    DEFAULT_REDFISH_CLIENT_MAX_RETRY,
+    DEFAULT_REDFISH_CLIENT_TIMEOUT,
+    DEFAULT_REDFISH_DISCOVER_CACHE_TTL,
+    Config,
+)
 from .exporter import Exporter
 
 logger = logging.getLogger(__name__)
@@ -112,6 +118,27 @@ def parse_command_line() -> argparse.Namespace:
         help="Enable redfish collector (default: disabled)",
         action="store_true",
     )
+    parser.add_argument(
+        "--redfish-client-timeout",
+        help="Redfish client timeout in seconds for initial connection (default: 3) ",
+        default=DEFAULT_REDFISH_CLIENT_TIMEOUT,
+        type=int,
+    )
+    parser.add_argument(
+        "--redfish-client-max-retry",
+        help="Number of times the redfish client will retry after timeout (default: 1) ",
+        default=DEFAULT_REDFISH_CLIENT_MAX_RETRY,
+        type=int,
+    )
+    parser.add_argument(
+        "--redfish-discover-cache-ttl",
+        help=(
+            "How long should the cached redfish discovery services "
+            "be stored in seconds (default: 86400) "
+        ),
+        default=DEFAULT_REDFISH_DISCOVER_CACHE_TTL,
+        type=int,
+    )
     args = parser.parse_args()
 
     return args
@@ -166,6 +193,9 @@ def main() -> None:
             redfish_username=namespace.redfish_username,
             redfish_password=namespace.redfish_password,
             ipmi_sel_interval=namespace.ipmi_sel_interval,
+            redfish_client_timeout=namespace.redfish_client_timeout,
+            redfish_client_max_retry=namespace.redfish_client_max_retry,
+            redfish_discover_cache_ttl=namespace.redfish_discover_cache_ttl,
         )
 
     # Start the exporter
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -4,9 +4,25 @@ from typing import Any, Dict, List, Optional
 
 import redfish
 import redfish_utilities
-from redfish.rest.v1 import HttpClient, InvalidCredentialsError, SessionCreationError
+from cachetools import TTLCache, cached
+from redfish.rest.v1 import (
+    HttpClient,
+    InvalidCredentialsError,
+    RetriesExhaustedError,
+    SessionCreationError,
+)
 
 logger = getLogger(__name__)
+
+# timeout in seconds for initial connection
+REDFISH_CLIENT_TIMEOUT = 3
+# Number of times a request will retry after a timeout
+REDFISH_CLIENT_MAX_RETRY = 1
+
+# maximum size of cache before old items are removed
+CACHE_MAXSIZE = 5
+# how long should the cache be stored in seconds
+CACHE_TTL = 86400
 
 
 class RedfishHelper:
@@ -44,12 +60,21 @@ class RedfishHelper:
         redfish_obj: Optional[HttpClient] = None
         try:
             redfish_obj = redfish.redfish_client(
-                base_url=host, username=username, password=password
+                base_url=host,
+                username=username,
+                password=password,
+                timeout=REDFISH_CLIENT_TIMEOUT,
+                max_retry=REDFISH_CLIENT_MAX_RETRY,
             )
             redfish_obj.login(auth="session")
             sensors = redfish_utilities.get_sensors(redfish_obj)
             return sensors
-        except (InvalidCredentialsError, SessionCreationError, ConnectionError) as err:
+        except (
+            InvalidCredentialsError,
+            SessionCreationError,
+            ConnectionError,
+            RetriesExhaustedError,
+        ) as err:
             logger.exception(err)
         finally:
             # Log out
@@ -57,8 +82,10 @@ class RedfishHelper:
                 redfish_obj.logout()
         return sensors
 
+    @cached(cache=TTLCache(maxsize=CACHE_MAXSIZE, ttl=CACHE_TTL))
     def discover(self) -> bool:
         """Return true if redfish is been discovered."""
+        logger.info("Discovering redfish services")
         services = redfish.discover_ssdp()
         if len(services) == 0:
             return False
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -1,10 +1,10 @@
 """Redfish collector."""
 from logging import getLogger
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 import redfish
 import redfish_utilities
-from cachetools import TTLCache, cached
+from cachetools.func import ttl_cache
 from redfish.rest.v1 import (
     HttpClient,
     InvalidCredentialsError,
@@ -14,27 +14,36 @@ from redfish.rest.v1 import (
 
 logger = getLogger(__name__)
 
-# timeout in seconds for initial connection
-REDFISH_CLIENT_TIMEOUT = 3
-# Number of times a request will retry after a timeout
-REDFISH_CLIENT_MAX_RETRY = 1
 
-# maximum size of cache before old items are removed
-CACHE_MAXSIZE = 5
-# how long should the cache be stored in seconds
-CACHE_TTL = 86400
+# pylint: disable=R0903
+# pylint: disable=R0913
 
 
 class RedfishHelper:
     """Helper function for redfish."""
 
-    def get_sensor_data(self, host: str, username: str, password: str) -> Dict[str, List]:
+    def __init__(self, discover_cache_ttl: int) -> None:
+        """Initialize disover method with TTL value."""
+        self.discover = self.get_discover(discover_cache_ttl)
+
+    def get_sensor_data(
+        self, host: str, username: str, password: str, timeout: int, max_retry: int
+    ) -> Dict[str, List]:
         """Get sensor data.
+
+        Params:
+            host: redfish URL
+            username: username to login
+            password: password to login
+            timeout: redfish client timeout
+            max_retry: redfish client max retry
 
         Returns:
             sensor_data: a dictionary where key, value maps to chassis name, sensor data.
         """
-        data = self._get_sensor_data(host=host, username=username, password=password)
+        data = self._get_sensor_data(
+            host=host, username=username, password=password, timeout=timeout, max_retry=max_retry
+        )
         if not data:
             return {}
         output = {}
@@ -54,8 +63,21 @@ class RedfishHelper:
             output[name] = sensors
         return output
 
-    def _get_sensor_data(self, host: str, username: str, password: str) -> Optional[List[Any]]:
-        """Return sensor if sensor exists else None."""
+    def _get_sensor_data(
+        self, host: str, username: str, password: str, timeout: int, max_retry: int
+    ) -> Optional[List[Any]]:
+        """Return sensor if sensor exists else None.
+
+        Params:
+            host: redfish URL
+            username: username to login
+            password: password to login
+            timeout: redfish client timeout
+            max_retry: redfish client max retry
+
+        Returns:
+            sensors: List of dicts with details for each sensor
+        """
         sensors: Optional[List[Any]] = None
         redfish_obj: Optional[HttpClient] = None
         try:
@@ -63,8 +85,8 @@ class RedfishHelper:
                 base_url=host,
                 username=username,
                 password=password,
-                timeout=REDFISH_CLIENT_TIMEOUT,
-                max_retry=REDFISH_CLIENT_MAX_RETRY,
+                timeout=timeout,
+                max_retry=max_retry,
             )
             redfish_obj.login(auth="session")
             sensors = redfish_utilities.get_sensors(redfish_obj)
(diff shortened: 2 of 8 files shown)


dashmage:
Not able to run `tox -e lint`. Raised another issue -- #35

rgildein (commented):
Can we provide simple unit tests to ensure that cache is working as we expected? Also there are no tests for `_get_sensor_data` function and it's probably out of scope of this PR, but it would be nice if you could add it also.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -4,10 +4,26 @@
...
+    HttpClient,
+    InvalidCredentialsError,
+    RetriesExhaustedError,
+    SessionCreationError,
+)
 
 logger = getLogger(__name__)
 
+# timeout in seconds for initial connection
+REDFISH_CLIENT_TIMEOUT = 3
```
jneo8, line 18: **nitpick:** 

Instead of putting settings here, how about loading them from the config?
We can provide a default value also.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -44,21 +60,32 @@ def _get_sensor_data(self, host: str, username: str, password: str) -> Optional[
...
         finally:
             # Log out
             if redfish_obj:
                 redfish_obj.logout()
         return sensors
 
+    @cached(cache=TTLCache(maxsize=CACHE_MAXSIZE, ttl=CACHE_TTL))
     def discover(self) -> bool:
         """Return true if redfish is been discovered."""
+        logger.info("Discovering redfish services")
```
jneo8, line 88: **nitpick (non-blocking):** 

Can we also put log to show the discover result?

jneo8 (changes requested):

jneo8 (changes requested):
Just as @rgildein said. Please provide unit tests in the PR.

sudeephb (commented):
After adding unit tests, as already mentioned in other comments, everything else LGTM.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: I think that it's best we keep these configs directly in `redfish.py` since it would be easier to refer them and change it here rather than going to a different file. 

The configs for `sasircu.py` collector are also inside the same file.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Good idea. I'll also add another info log in case no redfish services have been discovered.

prometheus_hardware_exporter/collectors/redfish.py
jneo8: I would like to change the settings with config instead of the source code change. It will be useful when debugging or turning.

prometheus_hardware_exporter/__main__.py
```diff
@@ -112,6 +118,24 @@ def parse_command_line() -> argparse.Namespace:
...
+    )
+    parser.add_argument(
+        "--redfish-client-max-retry",
+        help="Number of time redfish client will retry after timeout (default: 1) ",
+        default=DEFAULT_REDFISH_CLIENT_MAX_RETRY,
+        type=int,
+    )
+    parser.add_argument(
+        "--redfish-discover-cache-ttl",
+        help="How long should the cache be stored in seconds (default: 86400) ",
```
sudeephb, line 135:
```suggestion
        help="How long should the redfish discovery info stored in cache, in seconds (default: 86400) ",
```

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -38,28 +48,55 @@ def get_sensor_data(self, host: str, username: str, password: str) -> Dict[str,
             output[name] = sensors
         return output
 
-    def _get_sensor_data(self, host: str, username: str, password: str) -> Optional[List[Any]]:
+    def _get_sensor_data(
+        self, login_info: dict, timeout: int, max_retry: int
+    ) -> Optional[List[Any]]:
         """Return sensor if sensor exists else None."""
```
sudeephb, line 54: Can you also describe the parameters(especially login_info) in the docstring? Since it's now a dict, it's not very intuitive.

sudeephb (commented):
Few nitpick comments. Otherwise LGTM.

prometheus_hardware_exporter/__main__.py
dashmage: Done!

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Good suggestion, adding for both the sensor methods.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -1,24 +1,41 @@
...
+
+
 class RedfishHelper:
     """Helper function for redfish."""
+    def __init__(self, discover_cache_ttl: int) -> None:
+        """Initialize disover method with TTL value."""
+        self.discover = get_discover(discover_cache_ttl)
 
-    def get_sensor_data(self, host: str, username: str, password: str) -> Dict[str, List]:
+    def get_sensor_data(self, login_info: dict, timeout: int, max_retry: int) -> Dict[str, List]:
```
rgildein, line 28: not-blocking: I think it would be more readable to leave `host`, `username` and `password ` as arguments, instead of hiding it to one single argument.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -38,28 +56,61 @@ def get_sensor_data(self, host: str, username: str, password: str) -> Dict[str,
...
+        """Return true if redfish services have been discovered."""
+        logger.info("Discovering redfish services...")
         services = redfish.discover_ssdp()
         if len(services) == 0:
+            logger.info("No redfish services discovered")
             return False
+        logger.info("Discovered redfish services: %s", services)
         return True
+
+    return _discover
```
rgildein, line 116: Why we moved this out of the class? Why not:
```python
def discover(self) -> bool:
    @ttl_cache(ttl=self.ttl)
    def _discover() -> bool:
        ...

    return _discover()
```

I do not understand why we are always creating class object, which ware basically storing no values. before you add `__init__` there was no class variable, no variable defined in __init__ and no properties.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: The reason behind this was because there was a pylint error with too many arguments for the method (6 including `self`, it should be below 5). So I thought I'd collect all the redfish login related vars into one dict and pass that to the method.

Should I disable the pylint error instead and pass the vars separately?

prometheus_hardware_exporter/collectors/redfish.py
jneo8: The example somehow make the `@ttl_cache` not working. The purpose here is just let the decorator can get the config value as the ttl settings.

prometheus_hardware_exporter/collectors/redfish.py
jneo8: I am agree with Robert, `login_info` doesn't make things clear. We can try to ignore pylint here.

prometheus_hardware_exporter/collectors/redfish.py
rgildein: I think we can easily do:
```python
@ttl_cache(ttl=TTF)
def discover(self) -> bool:
...
```
or completely remove this class object and define just simple functions outside of it.

My issue is that we define class and them moved function, which is 100% related with the class, outside of class object only to override bad design of cachetools and our. I do not think that this is blocking, but it's just bad design.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Since both of you are suggesting this, I'll revert my change, separate all the args again and add a "pylint disable" comment. IMO I think it's best to keep the number of arguments low to avoid too much complexity in the function and to bundle related arguments in a data structure so that it's easier to understand, maintain and test overall.

@rgildein - IMO it's good to keep the default number of max arguments (6) and not ignore this error by default from the template. Ideally 3-4 args would be good for a function and if it needs more, it's probably a sign that the function is doing too much work and we can split into another function. For this specific case, I can add the comment to disable to disable the pylint error.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Thanks for the comment @rgildein !

The main aim here is to configure the ttl value for the cache at runtime which is coming from the the CLI args defined in `__main__.py`. I initially defined a class attribute which could then be modified at runtime to change the ttl value. But that didn't work since the decorator caches the initial value of the ttl class attribute when the class is initialized and doesn't change even when the attribute it modified later.

After some help from @jneo8 I defined a sort of wrapper function that returns the cached discover method with the ttl value being passed at runtime. 

I don't know why I didn't try this out earlier, but having the method inside the class itself seems to work while testing on a machine and it passes the unit tests as well. So we don't have to have the bad design after all :)

```python
class RedfishHelper:
    """Helper function for redfish."""

    def __init__(self, discover_cache_ttl: int) -> None:
        """Initialize disover method with TTL value."""
        self.discover = self.get_discover(discover_cache_ttl)

    (...)

    def get_discover(self, ttl: int) -> Callable:
        """Return the cached discover function.

        Passes the ttl value to the cache decorator at runtime.
        """

        @ttl_cache(ttl=ttl)
        def _discover() -> bool:
            """Return true if redfish services have been discovered."""
            logger.info("Discovering redfish services...")
            services = redfish.discover_ssdp()
            if len(services) == 0:
                logger.info("No redfish services discovered")
                return False
            logger.info("Discovered redfish services: %s", services)
            return True

        return _discover

**Review by jneo8 — APPROVED** [review]
LGTM

File: prometheus_hardware_exporter/collectors/redfish.py (hunk 5 above)
Comment (rgildein) [reply prometheus_hardware_exporter/collectors/redfish.py]: > IMO I think it's best to keep the number of arguments low to avoid too much complexity in the function and to bundle related arguments in a data structure so that it's easier to understand, maintain and test overall.

If you wrap 3 arguments into single one and than unwrap it and use all those arguments + other in single call, it's not cleaner, easier to tests and deferentially not providing lower complexity.

```python
def foo(a: Dict[str, Any], b: int, c: int) -> None:
    a1 = a["a1"]
    a2 = a["a2"]
    a3 = a["a3"]
    foo2(a1, a2, a3, b, c)
```

I'm not sure that 6 is always enough, since we are using a lot 3-th party staff and our functions sometimes needs too many arguments, because they are forwarding them to the external function. I think that 9 will make much more sense. I'm doing lot of reviews and everyday I'm saying usage of pylint disable for this.

Btw, never use force push after review, because now I could not check what did you changed from my last review. Thanks

prometheus_hardware_exporter/collectors/redfish.py
rgildein: Like this we are doing wrapper for wrapper, which is actually also wrapper. That is why I was trying to avoid using `get_discover`. But if there is no other solution, it's ok.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -1,24 +1,49 @@
...
+    InvalidCredentialsError,
+    RetriesExhaustedError,
+    SessionCreationError,
+)
 
 logger = getLogger(__name__)
 
 
+# pylint: disable=R0903
+# pylint: disable=R0913
```
rgildein, line 19: not-blocking: I think you can use `R0903, R0913`

rgildein (approved):
LGTM

prometheus_hardware_exporter/collectors/redfish.py
esunar: can you use extended version of those so we can understand without checking what it means
`# pylint: disable=too-few-public-methods`

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -1,24 +1,49 @@
...
+
+
 class RedfishHelper:
     """Helper function for redfish."""
 
-    def get_sensor_data(self, host: str, username: str, password: str) -> Dict[str, List]:
+    def __init__(self, discover_cache_ttl: int) -> None:
+        """Initialize disover method with TTL value."""
+        self.discover = self.get_discover(discover_cache_ttl)
```
esunar, line 27: Why don't we initialize the helper with host: str, username: str, password: str, timeout: int, max_retry: int

esunar: Or just with Config

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -38,28 +63,62 @@ def get_sensor_data(self, host: str, username: str, password: str) -> Dict[str,
             output[name] = sensors
         return output
 
-    def _get_sensor_data(self, host: str, username: str, password: str) -> Optional[List[Any]]:
-        """Return sensor if sensor exists else None."""
+    def _get_sensor_data(
```
esunar, line 66: two methods with same name?

esunar (changes requested):
I've some inline comments and questions. Can you check please

prometheus_hardware_exporter/collectors/redfish.py
rgildein: Private and public?

prometheus_hardware_exporter/collectors/redfish.py
dashmage: I understand the situation in your example where encapsulating all the args into a single one might not always have lesser complexity when it's easier to pass them all along separately. 

But I think as a general practice if the function is taking that many number of arguments (>6), it would be a bit complex to make sure to pass all the right args to the function each time you call it. Even during testing, we need to send all these args. But at the end of the day, it all depends on the exact situation :)

> Btw, never use force push after review, because now I could not check what did you changed from my last review.

Ah, it's a force of habit my bad. I don't want to mess up the log with lots of useless comments but I also understand we can just squash them before merging. Will make sure to not force push from now :+1:

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Yeah I really couldn't think of a more elegant solution to this problem either :disappointed:

prometheus_hardware_exporter/collectors/redfish.py
dashmage: It also looks like I can omit `RO903` for too-few-public-methods since the `get_discover` method is added in the class now.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Yep, `_get_sensor_data` is internal to the class while `get_sensor_data` actually returns data to the outside.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: That's a really nice idea. We wouldn't need to pass these login parameters each time to the get_sensor methods and while adding new methods for additional redfish metrics, their arg list would be simpler too. Will implement and push out this change :100:

prometheus_hardware_exporter/collectors/redfish.py
rgildein: > I understand the situation in your example where encapsulating all the args into a single one might not always have lesser complexity when it's easier to pass them all along separately.

I will not have any issue if you do 
```python
def foo(foo_arguments) -> None:
    """...
    params:
        foo_arguments: arguments for foo2
    """
    foo2(**foo_arguments)
```
Or simply
```python
def foo(*args, **kwargs) -> None:
    ...
    foo2(*args, **kwargs)
```

But in your case, when you never used `login_info` anywhere and simply pass those along with `timeout` and `max_retry` to sub-function, same function. Like this you only creates potential issue that someone will define `login_info` wrongly and got simple KeyError instead of `TypeError: 'a' is an invalid keyword argument for redfish.redfish_client`.

When you are using good IDE, it will tells you that you are missing e.g. `host` or that you provide wrong type for `host`, but when you do `login_info` it will tells you nothing.

dashmage:
Implemented very helpful changes suggested by @esunar . Also fixed the unit tests and added another one for the new method being implemented to map the sensor data.

esunar (approved):

prometheus_hardware_exporter/collectors/redfish.py
dashmage: After the latest changes, all the values are being sent to `RedfishHelper` as an instance of the `Config` object containing the login args (host, username, password) and the ttl value for the cache. This eliminates the need to send it all separately and also makes sure there is a tight coupling between the related data elements. It's also easier to instantiate the class now and prevents declaring an arbitrary `login_info` which isn't being used anywhere else.

rgildein (approved):
LGTM
I like idea with passing config to __init__ and now even class make much more sense :+1:. Good job.

PR #36, opened 2023-08-16
Add new redfish metrics

dashmage:
Add additional metrics for processor, memory, storage controllers, storage drives, smart storage, network adapters, chassis.

### DIFF
--- a/prometheus_hardware_exporter/collector.py
+++ b/prometheus_hardware_exporter/collector.py
@@ -1,7 +1,7 @@
 """Module for a collection of hardware collecters."""
 
 from logging import getLogger
-from typing import Dict, List
+from typing import Any, Dict, List
 
 from prometheus_client.metrics_core import GaugeMetricFamily, InfoMetricFamily
 
@@ -874,6 +874,61 @@ class RedfishCollector(BlockingCollector):
                 documentation="Sensor information obtained from redfish.",
                 metric_class=InfoMetricFamily,
             ),
+            Specification(
+                name="redfish_processors",
+                documentation="Total number of processors obtained from redfish.",
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="redfish_processor",
+                documentation="Processor information obtained from redfish.",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="redfish_storage_controllers",
+                documentation="Total number of storage controllers obtained from redfish.",
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="redfish_storage_controller",
+                documentation="Storage Controller information obtained from redfish.",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="redfish_network_adapters",
+                documentation="Total number of network adapters obtained from redfish.",
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="redfish_chassis",
+                documentation="Chassis information obtained from redfish.",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="redfish_storage_drives",
+                documentation="Total number of storage drives obtained from redfish.",
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="redfish_storage_drive",
+                documentation="Storage information obtained from redfish.",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="redfish_memory_dimms",
+                documentation="Total memory DIMMs obtained from redfish.",
+                metric_class=GaugeMetricFamily,
+            ),
+            Specification(
+                name="redfish_memory_dimm",
+                documentation="Memory DIMM information obtained from redfish.",
+                metric_class=InfoMetricFamily,
+            ),
+            Specification(
+                name="redfish_smart_storage_health",
+                documentation="Indicates if Smart Storage (if present) is healthy.",
+                metric_class=GaugeMetricFamily,
+            ),
         ]
 
     def fetch(self) -> List[Payload]:
@@ -882,13 +937,64 @@ class RedfishCollector(BlockingCollector):
         service_status = self.redfish_helper.discover()
         payloads.append(Payload(name="redfish_service_available", value=float(service_status)))
 
-        sensor_data = self.redfish_helper.get_sensor_data()
+        processor_count: Dict[str, int]
+        processor_data: Dict[str, List]
+        storage_controller_count: Dict[str, int]
+        storage_controller_data: Dict[str, List]
+        storage_drive_count: Dict[str, int]
+        storage_drive_data: Dict[str, List]
+        memory_dimm_count: Dict[str, int]
+        memory_dimm_data: Dict[str, List]
+
+        sensor_data: Dict[str, List] = self.redfish_helper.get_sensor_data()
+        processor_count, processor_data = self.redfish_helper.get_processor_data()
+        (
+            storage_controller_count,
+            storage_controller_data,
+        ) = self.redfish_helper.get_storage_controller_data()
+        network_adapter_count: Dict[str, Any] = self.redfish_helper.get_network_adapter_data()
+        chassis_data: Dict[str, Dict] = self.redfish_helper.get_chassis_data()
+        storage_drive_count, storage_drive_data = self.redfish_helper.get_storage_drive_data()
+        memory_dimm_count, memory_dimm_data = self.redfish_helper.get_memory_dimm_data()
+        smart_storage_health_data: Dict[
+            str, Any
+        ] = self.redfish_helper.get_smart_storage_health_data()
+
         if not sensor_data:
             logger.error("Failed to get sensor data via redfish.")
             payloads.append(Payload(name="redfish_call_success", value=0.0))
             return payloads
+        if not processor_data:
+            logger.error("Failed to get processor data via redfish.")
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+        if not storage_controller_data:
+            logger.error("Failed to get storage controller data via redfish.")
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+        if not network_adapter_count:
+            logger.error("Failed to get network adapter data via redfish.")
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+        if not chassis_data:
+            logger.error("Failed to get chassis data via redfish.")
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+        if not storage_drive_data:
+            logger.error("Failed to get storage drive data via redfish.")
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+        if not memory_dimm_data:
+            logger.error("Failed to get memory DIMM data via redfish.")
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+
+        if not smart_storage_health_data:
+            logger.info("Failed to get smart storage health data via redfish.")
 
         payloads.append(Payload(name="redfish_call_success", value=1.0))
+
+        # Appending sensor data to payload
         for chassis_name, curr_sensor_data in sensor_data.items():
             for sensor_data_item in curr_sensor_data:
                 payloads.append(
@@ -903,6 +1009,116 @@ class RedfishCollector(BlockingCollector):
                     )
                 )
 
+        # Appending processor data to payload
+        for system_id, curr_processor_data in processor_data.items():
+            payloads.append(
+                Payload(
+                    name="redfish_processors",
+                    value=processor_count[system_id],
+                    labels=[system_id],
+                )
+            )
+            for processor in curr_processor_data:
+                payloads.append(
+                    Payload(
+                        name="redfish_processor",
+                        value={
+                            "system_id": system_id,
+                            "processor_id": processor["processor_id"],
+                            "model": processor["model"],
+                            "health": processor["health"],
+                            "state": processor["state"],
+                        },
+                    )
+                )
+        # Appending storage controller data to payload
+        for system_id, curr_storage_controller_data in storage_controller_data.items():
+            payloads.append(
+                Payload(
+                    name="redfish_storage_controllers",
+                    value=int(storage_controller_count[system_id]),
+                    labels=[system_id],
+                )
+            )
+            for controller in curr_storage_controller_data:
+                payloads.append(
+                    Payload(
+                        name="redfish_storage_controller",
+                        value={
+                            "system_id": system_id,
+                            "storage_id": controller["storage_id"],
+                            "controller_id": controller["controller_id"],
+                            "health": controller["health"],
+                            "state": controller["state"],
+                        },
+                    )
+                )
+
+        # Appending chassis data to payload
+        for chassis_id, count in network_adapter_count.items():
+            payloads.append(
+                Payload(
+                    name="redfish_network_adapters",
+                    value=int(count),
+                    labels=[chassis_id],
+                )
+            )
+
+        # Appending storage drive data to payload
+        for system_id, curr_storage_drive_data in storage_drive_data.items():
+            payloads.append(
+                Payload(
+                    name="redfish_storage_drives",
+                    value=int(storage_drive_count[system_id]),
+                    labels=[system_id],
+                )
+            )
+            for drive in curr_storage_drive_data:
+                payloads.append(
+                    Payload(
+                        name="redfish_storage_drive",
+                        value={
+                            "system_id": system_id,
+                            "storage_id": drive["storage_id"],
+                            "drive_id": drive["drive_id"],
+                            "health": drive["health"],
+                            "state": drive["state"],
+                        },
+                    )
+                )
+
+        # Appending memory DIMM data to payload
+        for system_id, curr_memory_dimm_data in memory_dimm_data.items():
+            payloads.append(
+                Payload(
+                    name="redfish_memory_dimms",
+                    value=int(memory_dimm_count[system_id]),
+                    labels=[system_id],
+                )
+            )
+            for memory_dimm in curr_memory_dimm_data:
+                payloads.append(
+                    Payload(
+                        name="redfish_memory_dimm",
+                        value={
+                            "system_id": system_id,
+                            "memory_id": memory_dimm["memory_id"],
+                            "health": memory_dimm["health"],
+                            "state": memory_dimm["state"],
+                        },
+                    )
+                )
+
+        # Appending smart storage health data to payload (if present)
+        if smart_storage_health_data:
+            for chassis_id, curr_smart_storage_health_data in smart_storage_health_data.items():
+                payloads.append(
+                    Payload(
+                        name="redfish_smart_storage_health",
+                        value=1.0 if curr_smart_storage_health_data["health"] == "OK" else 0.0,
+                        labels=[chassis_id],
+                    )
+                )
         return payloads
 
     def process(self, payloads: List[Payload], datastore: Dict[str, Payload]) -> List[Payload]:
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -1,12 +1,14 @@
 """Redfish collector."""
 from logging import getLogger
-from typing import Any, Callable, Dict, List
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import redfish
 import redfish_utilities
 from cachetools.func import ttl_cache
 from redfish.rest.v1 import (
+    HttpClient,
     InvalidCredentialsError,
+    RestResponse,
     RetriesExhaustedError,
     SessionCreationError,
 )
@@ -15,18 +17,45 @@ from prometheus_hardware_exporter.config import Config
 
 logger = getLogger(__name__)
 
+# pylint: disable=too-many-instance-attributes
+
 
 class RedfishHelper:
     """Helper function for redfish."""
 
     def __init__(self, config: Config) -> None:
-        """Initialize redfish login args and cache TTL value for discover method."""
+        """Initialize values for class"""
         self.host = config.redfish_host
         self.username = config.redfish_username
         self.password = config.redfish_password
         self.timeout = config.redfish_client_timeout
         self.max_retry = config.redfish_client_max_retry
         self.discover = self.get_discover(config.redfish_discover_cache_ttl)
+
+        self.sys_inventory: Optional[List[Dict[str, Any]]] = None
+        self.redfish_obj: Optional[HttpClient] = None
+
+    def _get_redfish_obj(self) -> Optional[HttpClient]:
+        """Return a new redfish object."""
+        try:
+            self.redfish_obj = redfish.redfish_client(
+                base_url=self.host,
+                username=self.username,
+                password=self.password,
+                timeout=self.timeout,
+                max_retry=self.max_retry,
+            )
+            self.redfish_obj.login(auth="session")
+            logger.info("Logging into redfish service...")
+        except (
+            InvalidCredentialsError,
+            SessionCreationError,
+            ConnectionError,
+            RetriesExhaustedError,
+        ) as err:
+            logger.exception(err)
+
+        return self.redfish_obj
 
     def get_sensor_data(self) -> Dict[str, List]:
         """Get sensor data.
@@ -38,6 +67,7 @@ class RedfishHelper:
         return self._map_sensor_data_to_chassis(data)
 
     def _map_sensor_data_to_chassis(self, sensor_data: List[Any]) -> Dict[str, List]:
+        """Return dictionary with sensor data mapped to chassis."""
         output = {}
         for chassis in sensor_data:
             output[str(chassis["ChassisName"])] = [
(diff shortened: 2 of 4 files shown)


dashmage:
Pending:
- Some unit tests (for failure cases)
- Redfish exporter takes too long to query now (~10s). Need to speed it up (redfish object login/logout once, increase scrape timeout, run in parallel)
- Pylint errors due to large amount of code in the `RedfishCollector` class in `collector.py`. Might need some refactoring.
```
pylint --recursive=y .
************* Module prometheus_hardware_exporter.collector
prometheus_hardware_exporter/collector.py:1:0: C0302: Too many lines in module (1126/1000) (too-many-lines)
prometheus_hardware_exporter/collector.py:934:4: R0914: Too many local variables (30/15) (too-many-locals)
prometheus_hardware_exporter/collector.py:934:4: R0911: Too many return statements (8/6) (too-many-return-statements)
prometheus_hardware_exporter/collector.py:934:4: R0912: Too many branches (21/12) (too-many-branches)
prometheus_hardware_exporter/collector.py:934:4: R0915: Too many statements (56/50) (too-many-statements)
```

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -15,19 +17,46 @@
...
         self.username = config.redfish_username
         self.password = config.redfish_password
         self.timeout = config.redfish_client_timeout
         self.max_retry = config.redfish_client_max_retry
         self.discover = self.get_discover(config.redfish_discover_cache_ttl)
 
+        self.sys_inventory: Optional[List[Dict[str, Any]]] = None
+        self.redfish_obj: Optional[HttpClient] = None
+
+    def _get_redfish_obj(self) -> Optional[HttpClient]:
```
jneo8, line 38: **issue:** 

Return None is not a nice practice of error handling.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,23 +86,338 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
-            InvalidCredentialsError,
-            SessionCreationError,
-            ConnectionError,
-            RetriesExhaustedError,
-        ) as err:
-            logger.exception(err)
-        return []
+        sensors: List[Any] = []
+
+        redfish_obj: Optional[HttpClient] = self._get_redfish_obj()
```
jneo8, line 91: **issue (blocking):**  

I saw this `_get_redfish_obj` in multiple different fetch data function.
I would like to know how long this function takes. Because multiple times login/logout maybe cause performance issue here.

Also I would like to keep the login session for later use. Can we use some similar design pattern from web crawler?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,23 +86,338 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+    def _verify_redfish_call(self, redfish_obj: HttpClient, uri: str) -> Optional[Dict[str, Any]]:
+        """Return REST response for GET request API call on the URI.
+
+        Returns None if URI isn't present.
+        """
+        resp: RestResponse = redfish_obj.get(uri)
+        if resp.status == 200:
+            return resp.dict
+        logger.info("Not able to query from URI: %s.", uri)
+        return None
```
jneo8, line 107: **nitpick:** 

We don't need this line.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,23 +86,338 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+            redfish_obj.logout()
+        return sensors
+
+    def _verify_redfish_call(self, redfish_obj: HttpClient, uri: str) -> Optional[Dict[str, Any]]:
+        """Return REST response for GET request API call on the URI.
+
+        Returns None if URI isn't present.
+        """
+        resp: RestResponse = redfish_obj.get(uri)
+        if resp.status == 200:
```
jneo8, line 104: **praise (non-blocking):** 

Instead of using `200`, can we use some `HttpStatusXXX` variable from standard library?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,23 +86,338 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+        logger.info("Not able to query from URI: %s.", uri)
+        return None
+
+    def get_processor_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
+        """Return processor data and count.
+
+        Returns:
+            processor_count: Dict of system ids mapped to the total number
+            of processors on that system.
+            processor_data: Dict of system ids mapped to list of processors on that system.
```
jneo8, line 115: **suggestion:** 

We should provide the example of processor_data in doc-string.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,23 +86,338 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+                                "state": data["Status"]["State"],
+                                "health": data["Status"]["Health"] or "NA",
+                            }
+                        )
+
+                storage_controller_data[system_id] = storage_controller_data_in_curr_system
+
+            redfish_obj.logout()
+
+        logger.info("storage controller count: %s", storage_controller_count)
```
jneo8, line 214: **nitpick:** 

This shouldn't be info level log.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,23 +86,338 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+                                "storage_id": storage_id,
+                                "controller_id": data["MemberId"],
+                                "state": data["Status"]["State"],
+                                "health": data["Status"]["Health"] or "NA",
+                            }
+                        )
+
+                storage_controller_data[system_id] = storage_controller_data_in_curr_system
+
+            redfish_obj.logout()
```
jneo8, line 212: **nitpick (blocking):** 

What if error happen and this line is not been called? We should change the practice here.

jneo8 (changes requested):
Please see my inline reviews.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Here is a [pastebin link](https://pastebin.ubuntu.com/p/tmwhJB6cJQ/) to the full output when querying from the exporter using only the redfish collector.

- The average time for running `curl localhost:10000` currently is about 11s. 
- Creating a new REST session and logging in takes about 0.8s. Logging out takes about 0.1s. So the whole process would be around 1s.
- We're creating a new redfish object and performing login/logout for each metric, or 8 times or about 8s. If we keep only one session to use for each metric and login/logout only once, we can save 7s here.
- Most of the other API calls take < 0.1s

I would need to see how I can instantiate a new redfish object once when the `RedfishHelper` class object is created  and then logout once this object is destroyed.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Which line are you referring to here? Is it `return None` in which case I'm not sure what should we return. Or do you mean the `logger.info` statement? If the method which is calling `_verify_redfish_call` prints a descriptive log message in the case where we return `None`, then I guess we wouldn't need to log here.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Will change it to debug level log :+1:

Pjack (changes requested):
I see redfish_obj.logout() in each function. My god feeling there is enhance space here. James also raises this topic.

prometheus_hardware_exporter/collector.py
```diff
@@ -882,13 +937,64 @@ def fetch(self) -> List[Payload]:
...
+        memory_dimm_count, memory_dimm_data = self.redfish_helper.get_memory_dimm_data()
+        smart_storage_health_data: Dict[
+            str, Any
+        ] = self.redfish_helper.get_smart_storage_health_data()
+
         if not sensor_data:
             logger.error("Failed to get sensor data via redfish.")
             payloads.append(Payload(name="redfish_call_success", value=0.0))
             return payloads
+        if not processor_data:
```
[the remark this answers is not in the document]
dashmage: I agree with your points. I will make both changes:
- Consolidate all <metric>_data into a data structure and then check in a loop
- Do not set `redfish_call_success` to 0 if just one (or few) metrics aren't available.

prometheus_hardware_exporter/collectors/redfish.py
jneo8: So we can ask some questions here:

- How long can I keep this session open?
- When should I create the session?
  - Should I create the session when the service start?
  - Should I create the session when I receive the request?
- Where should I store this session?
- Is there any pattern that can help me to get the same session every time when I send the request to redfish endpoint?

prometheus_hardware_exporter/collectors/redfish.py
dashmage: After looking into this for a bit, I have a different opinion now.

On checking code in the redfish library, it looks like they are using the status codes directly without `HTTPStatus` as you can see [here](https://github.com/DMTF/python-redfish-library/blob/main/src/redfish/rest/v1.py#L1004) and in the rest of the module. I feel like we can follow the same approach.

esunar (changes requested):
I've some comments.

prometheus_hardware_exporter/collectors/redfish.py
jneo8: The line 107. It will return None even you don't add it.

dashmage: Ah okay, it completely slipped my mind that if a function doesn't return anything it effectively returns None. Thanks for the suggestion :)

prometheus_hardware_exporter/collectors/redfish.py
jneo8: The purpose are very different here.
In the original function, redfish_obj maybe none is because we use it to control if we need to logout in the end, in the same function.

But `_get_redfish_obj` doesn't include this logic, so user may want to know the actual error instead of just None.

prometheus_hardware_exporter/collector.py
```diff
@@ -903,6 +1009,116 @@ def fetch(self) -> List[Payload]:
                     )
                 )
 
+        # Appending processor data to payload
```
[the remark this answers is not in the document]
dashmage: Good idea. I'll move this code to another `_add_metrics_to_payload` method.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Update: Looks like I need to explicitly add the return statement. Otherwise, mypy throws an error -
`prometheus_hardware_exporter/collectors/redfish.py:98: error: Missing return statement  [return]`

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -15,19 +17,46 @@
...
+        """Return a new redfish object."""
+        try:
+            self.redfish_obj = redfish.redfish_client(
+                base_url=self.host,
+                username=self.username,
+                password=self.password,
+                timeout=self.timeout,
+                max_retry=self.max_retry,
+            )
+            self.redfish_obj.login(auth="session")
```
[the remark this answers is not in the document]
dashmage: Added context manager, so we don't need to use this line

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -15,19 +17,46 @@
...
         self.max_retry = config.redfish_client_max_retry
         self.discover = self.get_discover(config.redfish_discover_cache_ttl)
 
+        self.sys_inventory: Optional[List[Dict[str, Any]]] = None
+        self.redfish_obj: Optional[HttpClient] = None
+
+    def _get_redfish_obj(self) -> Optional[HttpClient]:
+        """Return a new redfish object."""
+        try:
+            self.redfish_obj = redfish.redfish_client(
```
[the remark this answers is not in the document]
dashmage: Yes, I've now added these methods for RedfishHelper class!

dashmage:
> I see redfish_obj.logout() in each function. My god feeling there is enhance space here. James also raises this topic.

Yes, I've now made `RedfishHelper` class as a context manager so that the `__exit__` method calls `logout`.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: This behaviour is now changed while making the RedfishHelper class into a context manager.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: > It's not different to use 200 or HTTPStatus. Just because HTTPStatus is more clear. https://docs.python.org/3/library/http.html

IMO 200 is already quite clear and it's even used in the main redfish library code. So, we could also follow that leave it at 200 since anyone can infer from the context about the meaning. 

But if you insist, I can change 200 --> `HTTPStatus.OK` which would be more verbose.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: This is now resolved since I've added `__enter__` and `__exit__` methods to RedfishHelper class to allow it to be used as a context manager. Now when the context manager exists, the `logout` method is called and the session is closed. Also, we only have to login/logout once now.

prometheus_hardware_exporter/collector.py
```diff
@@ -903,8 +1008,111 @@ def fetch(self) -> List[Payload]:
                     )
                 )
 
-        return payloads
+        for system_id, curr_processor_data in metrics["processor_data"].items():
```
esunar, line 1011: this is still too long. each for loop can have its own function with a meaningful name

esunar (changes requested):

prometheus_hardware_exporter/collector.py
dashmage: Hmm, alright. Let me create individual functions for adding each of the individual metrics to the payload.

prometheus_hardware_exporter/collector.py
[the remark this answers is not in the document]
Pjack: We should discuss if there is a basic check, (e.g. HEAD redfish url,  or setup the session only) 
If this checking fail => redfish_call_success to 0,  and all other requests can be skipped. 
That is the main purpose of this metrics.

You should take other collects' code as example
https://github.com/canonical/prometheus-hardware-exporter/blob/92f6264e1b1bdb69f8dd8cbe26fd7f6944c8af48/prometheus_hardware_exporter/collectors/perccli.py#L54

prometheus_hardware_exporter/collector.py
```diff
@@ -874,22 +874,127 @@ def specifications(self) -> List[Specification]:
...
+            "smart_storage_health_data": smart_storage_health_data,
+        }
+
+        for metric_name, metric_data in metrics.items():
+            if not metric_data:
+                logger.error("Failed to get %s via redfish", metric_name)
 
         payloads.append(Payload(name="redfish_call_success", value=1.0))
-        for chassis_name, curr_sensor_data in sensor_data.items():
+
```
Pjack, line 984: When will we set redfish_call_success as 0?

prometheus_hardware_exporter/collector.py
Pjack: I'm okay with this function. Although it's long, but the structure is simple, we add the payload one by one. It's not difficult to understand what we want to do in this function. But from unit test's point of view, it's may be easier to write unit test when we split into small function. I will let Ashley to make the final decision.

Pjack (changes requested):

dashmage:
> I'm okay with this function. Although it's long, but the structure is simple, we add the payload one by one. It's not difficult to understand what we want to do in this function. But from unit test's point of view, it's may be easier to write unit test when we split into small function.

I think it's better to split it into separate methods so the logic is very clear and each method performs only one thing. And it would be easier to write the unit tests as well as you have mentioned.

prometheus_hardware_exporter/collector.py
dashmage: That's a good question. We're setting `redfish_service_available` to 0 in case redfish discover method doesn't find any redfish services. But `redfish_call_success` is very vague. Which call does it refer to since there's a lot of API calls that we're making to redfish? Maybe we can check a few important metrics and if any of them doesn't return anything, we can set it to 0. But I'm also not 100% sure about this.

Pjack: Please provide some proposals.

prometheus_hardware_exporter/collector.py
```diff
@@ -874,21 +875,141 @@ def specifications(self) -> List[Specification]:
...
+            "smart_storage_health_data": smart_storage_health_data,
+        }
+
+        for metric_name, metric_data in metrics.items():
+            if not metric_data:
+                logger.error("Failed to get %s via redfish", metric_name)
 
         payloads.append(Payload(name="redfish_call_success", value=1.0))
+
+        payloads = self._create_sensor_metric_payload(metrics["sensor_data"], payloads)
```
Pjack, line 986: It's better to write it something like this
```suggestion
        payloads.extend (self._create_sensor_metric_payload(metrics["sensor_data"]))
```
It will decoupling of each internal function.

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
dashmage: This is a nice idea :+1:

prometheus_hardware_exporter/collector.py
dashmage: After putting some I think I got the right proposal for this.

`redfish_call_success` should be mapped to the result for the initial redfish session login. In case the login fails, there's no point calling any of the metrics. 

So I've added the code below so that in case the `_get_redfish_obj` method returns `None` (which happens if the login fails due to an exception), then `redfish_call_success` is 0.

```python
        with RedfishHelper(self.config) as redfish_helper:
            # redfish_obj returns None if login was not successful
            if redfish_helper.redfish_obj is None:
                payloads.append(Payload(name="redfish_call_success", value=0.0))
                return payloads
            
           (...)
```

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
```diff
@@ -936,6 +938,10 @@ def fetch(self) -> List[Payload]:
         """Load redfish data."""
         payloads: List[Payload] = []
         with RedfishHelper(self.config) as redfish_helper:
+            # redfish_obj returns None if login was not successful
+            if redfish_helper.redfish_obj is None:
+                payloads.append(Payload(name="redfish_call_success", value=0.0))
+                return payloads
```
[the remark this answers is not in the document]
dashmage: Good catch :+1:

prometheus_hardware_exporter/collector.py
```diff
@@ -936,6 +938,10 @@ def fetch(self) -> List[Payload]:
         """Load redfish data."""
         payloads: List[Payload] = []
         with RedfishHelper(self.config) as redfish_helper:
+            # redfish_obj returns None if login was not successful
+            if redfish_helper.redfish_obj is None:
```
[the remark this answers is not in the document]
dashmage: IMO `x is None` is more clearer and is a nicer check for `None` than `not x`. `not x` could be `True` when x is any boolean false value like `[]`, `""` etc. But `x is None` specifically checks whether the identity of `x` is `None` and isn't true  for any boolean false value. 

```python
>>> x = None
>>> if x is None:
...   print("x is None output")
... 
x is None output
>>> if not x:
...   print("not x output")
... 
not x output


>>> x = []
>>> if x is None:
...   print("x is None output")
... 
>>> if not x:
...   print("not x output")
... 
not x output
```

[pep8](https://peps.python.org/pep-0008/#programming-recommendations) recommends using `is` and `is not` as well:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.
Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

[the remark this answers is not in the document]
Pjack: When you must compare with "None", `is` and `is not` is better. (agree)

But I want to mention, it's not a good design if the reutrn value has different meaning between `[]` and `None`. In general case, we don't do that. It confuses the users. 
Unless these two cases are different, I prefer to change the syntax
* redfish_helper.redfish_obj = None
* redfish_helper.redfish_obj = []

Need @jneo8 to give 3rd opinion.

prometheus_hardware_exporter/collectors/redfish.py
Pjack: https://peps.python.org/pep-0008/#programming-recommendations
image

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,39 +115,424 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+        return sensors
+
+    def _verify_redfish_call(self, redfish_obj: HttpClient, uri: str) -> Optional[Dict[str, Any]]:
+        """Return REST response for GET request API call on the URI.
+
+        Returns None if URI isn't present.
+        """
+        resp: RestResponse = redfish_obj.get(uri)
+        if resp.status == 200:
+            return resp.dict
```
Pjack, line 135: Is this possible return empty dictionary? What doesn't that mean if it happen?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -304,8 +308,9 @@ def get_network_adapter_data(self) -> Dict[str, int]:
             network_adapters: Optional[Dict[str, Any]] = self._verify_redfish_call(
                 self.redfish_obj, network_adapters_root_uri
             )
-            if not network_adapters:
+            if network_adapters is None:
```
Pjack, line 311: Ditto, what if network_adapters is an empty dict? 
1. If None and "[]" has different meaning  => you should handle the empty dict case, but you didn't
2. If None and "[]" are identical =>  you should not compare with "None" only. 

Either one, you have to change the code. The original coding is good to me.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -527,9 +532,9 @@ def get_smart_storage_health_data(self) -> Dict[str, Any]:
             smart_storage_data: Optional[Dict[str, Any]] = self._verify_redfish_call(
                 self.redfish_obj, smart_storage_uri
             )
-            if not smart_storage_data:
+            if smart_storage_data is None:
```
Pjack, line 535: Ditto

Pjack (changes requested):

prometheus_hardware_exporter/collector.py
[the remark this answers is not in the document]
jneo8: For `if x is None` vs `if not x`, it's a standard practice in python to use `if x is None`.

I also mentioned in this [comment](https://github.com/canonical/prometheus-hardware-exporter/pull/36/files#r1295707578) about the bad practice of `_get_redfish_obj`.

But the most important part for me is that it's not make sense to check the connection inside the ContextManager, then why I need this ContextManager? It's not a good practice for me. Also you can see the discover function on top few lines of this line, it's second bad practice that try to connect to redfish but ignore the discover result.

I will suggest to re-design the ContextManager and discover function.

prometheus_hardware_exporter/collector.py
```diff
@@ -853,7 +856,10 @@ class RedfishCollector(BlockingCollector):
     def __init__(self, config: Config) -> None:
         """Initialize RedfishHelper instance."""
         super().__init__(config)
-        self.redfish_helper = RedfishHelper(config)
+        self.config = config
+        self.discover_redfish_services = get_cached_discover_function(
```
jneo8, line 860: **issue (blocking):** 

I don't get why we want to move discover to here. This somehow break the boundary of RedfishHelper and Collector.

The only path for Collector to get information from Redfish is through RedfishHelper.

jneo8 (changes requested):
There are some problem for RedfishHelper ContextManager and discover function design.

prometheus_hardware_exporter/collector.py
dashmage: I moved the discover method outside the RedfishHelper class because the caching was not working while using the context manager. Since the object of the class is instantiated newly each time we enter the context manager and then destroyed at the end, the caching doesn't work and the delay for discovering the redfish services is always present.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: It slipped my mind about the empty dict case. Will handle this case and keep the original code which will take care of None (from `_verify_redfish_call`) and empty dict

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Will revert to original code

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Yes, there could be a case where the URI is reachable (status is 200) but there's no elements present in it (empty dict). In my testing I haven't seen this case but it's good to also plan around this case.

prometheus_hardware_exporter/collector.py
dashmage: Made some changes to both the context manager and the discover method.

- The `__enter__` method in the context manager does not try to handle the exception which may be caused while getting the new redfish session. If an error occurs, the exception is raised and the caller will need to handle it.
In our case, this exception is captured and handled by the try...except statements wrapping the context manager in the RedfishCollector.fetch method.

- The discover method is now a static method inside the RedfishHelper class. This is used by the RedfishCollector class to discover redfish services and we can keep all the functionality within the RedfishHelper class itself. I've tested the cache and it works as well.

prometheus_hardware_exporter/collector.py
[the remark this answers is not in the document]
dashmage: Made some changes to both the context manager and the discover method.

- The `__enter__` method in the context manager does not try to handle the exception which may be caused while getting the new redfish session. If an error occurs, the exception is raised and the caller will need to handle it.
In our case, this exception is captured and handled by the try...except statements wrapping the context manager in the RedfishCollector.fetch method.

- The discover method is now a static method inside the RedfishHelper class. This is used by the RedfishCollector class to discover redfish services and we can keep all the functionality within the RedfishHelper class itself. I've tested the cache and it works as well.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -51,12 +48,18 @@ def __init__(self, config: Config) -> None:
         self.password = config.redfish_password
         self.timeout = config.redfish_client_timeout
         self.max_retry = config.redfish_client_max_retry
-        self.redfish_obj: HttpClient = self._get_redfish_obj()
+        self.redfish_obj: HttpClient = None
```
jneo8, line 48: **nitpick:** 

Missing typing

```
self.redfish_obj: Optional[HttpClient] = None
```

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -308,7 +290,7 @@ def get_network_adapter_data(self) -> Dict[str, int]:
             network_adapters: Optional[Dict[str, Any]] = self._verify_redfish_call(
                 self.redfish_obj, network_adapters_root_uri
             )
-            if network_adapters is None:
+            if not network_adapters:
                 logger.debug("No network adapters could be found on chassis id: %s", chassis_id)
                 print("here")
```
jneo8: **nitpick:** 

Remove print.

prometheus_hardware_exporter/collector.py
```diff
@@ -874,21 +885,160 @@ def specifications(self) -> List[Specification]:
...
+            "chassis_data": chassis_data,
+            "storage_drive_data": storage_drive_data,
+            "storage_drive_count": storage_drive_count,
+            "memory_dimm_data": memory_dimm_data,
+            "memory_dimm_count": memory_dimm_count,
+            "smart_storage_health_data": smart_storage_health_data,
+        }
+        for metric_name, metric_data in metrics.items():
+            if not metric_data:
+                logger.error("Failed to get %s via redfish", metric_name)
```
jneo8, line 1007: **nitpick:** 

The logging level should not be error.

tests/unit/test_collector.py
```diff
@@ -546,22 +547,100 @@ def test_105_perccli_physical_device_command_success(self):
...
-        redfish_collector.redfish_helper.get_sensor_data.return_value = {
+
+        mock_redfish_client.side_effect = ConnectionError
+        payloads = redfish_collector.collect()
+
+        self.assertEqual(len(list(payloads)), 2)
+        mock_redfish_client.assert_called_once()
+
+    @patch("prometheus_hardware_exporter.collector.logger")
+    @patch(
```
jneo8, line 599: **nitpick (blocking):** 

Maybe one patch for RedfishHelper is enough. As I understand from the test function name, the purpose for this test should be Collector, so mock RedfishHelper should be the correct way.

tests/unit/test_redfish.py
```diff
@@ -2,19 +2,108 @@
...
+                username="",
+                password="",
+                timeout=10,
+                max_retry=5,
+            )
+            mock_redfish_login.assert_called_once_with(auth="session")
+        mock_redfish_logout.assert_called_once()
+
+    @patch("prometheus_hardware_exporter.collectors.redfish.redfish.redfish_client")
+    def test_01_redfish_helper_context_manager_fail(self, mock_redfish_client):
```
jneo8, line 45: **nitpick (blocking):** 

The purpose of this test function maybe need to redefine?
We are not testing any error handling here.

We maybe want to test the error handling in collector?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -56,39 +98,424 @@ def _retrieve_redfish_sensor_data(self) -> List[Any]:
...
+            ],
+        }
 
-        Passes the ttl value to the cache decorator at runtime.
         """
+        storage_controller_data: Dict[str, List] = {}
+        # storage controllers on each system
+        storage_controller_count: Dict[str, int] = {}
+        storage_root_uri_pattern = "/redfish/v1/Systems/{}/Storage"
+        logger.info("Getting storage controller data...")
```
jneo8, line 226: **nitpick:** 

Debug?

prometheus_hardware_exporter/collector.py
```diff
@@ -902,7 +1052,174 @@ def fetch(self) -> List[Payload]:
                         },
                     )
                 )
+        return payloads
+
+    def _create_processor_metric_payload(
```
jneo8, line 1057: **nitpick:** 

The doc-string practice is not very good for those `_create_xxx` functions. It's worth to provide some example if the struct is complex.

jneo8 (changes requested):
- Please check the log level in the PR.
- Please provide some helping message in doc-string for `_create_xxx` functions.

Other please see my inline comments.

prometheus_hardware_exporter/collector.py
jneo8: Create redfish discover service here still not a good practice.
But the solution for this maybe separate the ContextManager and RedfishHelper.

I will put that in refactoring task instead of here.

Pjack (approved):
LGTM

tests/unit/test_collector.py
```diff
@@ -546,22 +547,100 @@ def test_105_perccli_physical_device_command_success(self):
...
+        mock_get_cached_discover_method.side_effect = cached_discover
         redfish_collector = RedfishCollector(Mock())
         redfish_collector.redfish_helper = Mock()
-        redfish_collector.redfish_helper.discover.return_value = False
-        redfish_collector.redfish_helper.get_sensor_data.return_value = {
+
+        mock_redfish_client.side_effect = ConnectionError
+        payloads = redfish_collector.collect()
+
+        self.assertEqual(len(list(payloads)), 2)
```
Pjack, line 595: It's not enough to only compare the length. What don't we list down the expected metrics explicitly? However, I find the same practice in sas2/sas3/... other interface in this file. 

Therefore, please at least to add comment to explain why the length should be 2. thanks!

Pjack (changes requested):
nitpick

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Good catch, I've fixed it :)

prometheus_hardware_exporter/collectors/redfish.py
dashmage: I caught this one already but it shouldn't have gone into the commit in the first place. Thanks for pointing it out though!

prometheus_hardware_exporter/collector.py
dashmage: Changed it to warning

prometheus_hardware_exporter/collectors/redfish.py
dashmage: IMO, it  should be an INFO level log.

Looking at [this doc page on logging](https://docs.python.org/3/library/logging.html#logging.INFO), it mentions:
- INFO: Confirmation that things are working as expected.
- DEBUG: Detailed information, typically only of interest to a developer trying to diagnose a problem.

So in this case, we are confirming that the storage controller data is being obtained and working correctly. 

What are your thoughts?

prometheus_hardware_exporter/collector.py
dashmage: Agree with you here. I'll add some helpful docstrings :+1:

prometheus_hardware_exporter/collectors/redfish.py
dashmage: After thinking about it a bit, I've modified the code like so:
```python
# redfish.py

    def __init__(self, config: Config) -> None:
        """Initialize values for class."""
        self.host = config.redfish_host
        self.username = config.redfish_username
        self.password = config.redfish_password
        self.redfish_obj: HttpClient
```

I'm not setting an initial value for `redfish_obj` now and setting the type only to `HttpClient`. Because in case the redfish_client login fails, we're anyway throwing an exception and raising that error.

What are your thoughts on this?

prometheus_hardware_exporter/collectors/redfish.py
jneo8: Because not all the log which show this `Getting ... data` has the same level. I would like to unified them.

dashmage: Currently, all the logs for `Getting <metric> data` are the same INFO level if you check.

tests/unit/test_collector.py
dashmage: I've patched the `__enter__` and `__exit__` methods now since the RedfishHelper is used as a context manager inside RedfishCollector. I've removed the extra patching of all the other methods in RedfishHelper.

tests/unit/test_redfish.py
dashmage: We are testing whether the right exception is being raised if there's an error while getting the redfish object at the start of the context manager execution:

```python
# test_redfish.py

for err in [
    InvalidCredentialsError(),
    ConnectionError(),
    SessionCreationError(),
    RetriesExhaustedError(),
]:
    mock_redfish_client.side_effect = err
    print(type(err))
    with self.assertRaises(
        (
            InvalidCredentialsError,
            ConnectionError,
            SessionCreationError,
            RetriesExhaustedError,
        )
    ):
        with RedfishHelper(mock_config):
            pass
```

This code loops through each of the possible exceptions and then opens the RedfishHelper context manager. We are asserting that each of these exception are being raised (from the `__enter__` method of RedfishHelper) using the outer context manager for `self.assertRaises`.

jneo8: You have this line`mock_redfish_client.side_effect = err`. Of course it will raise the error.

You should test the error handling try catch instead of this.

dashmage: The line `mock_redfish_client.side_effect = err` is mocking the exception only for the redfish_client. I'm testing whether that it is being raised during the usage of the context manager for RedfishHelper also. 

> You should test the error handling try catch instead of this.
I'm also testing that in test_collector.py but only for ConnectionError. I'll add the other exceptions as well there.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: This has been fixed by having the exception be raised directly from redfish_client in the `__enter__` method of RedfishHelper. This is caught while using the context manager in RedfishCollector.

esunar (approved):

PR #38, opened 2023-10-03
refactor: Replace redfish discover method.

dashmage:
* redfish.discover_ssdp method is replaced with a simple login call to the redfish service. If the RetriesExhaustedError exception is raised, we can infer that the redfish service isn't present on the system.
* Caching functionality of redfish discover result is maintained.
* Imports redfish_client method directly instead of importing entire redfish module.
* Modifies unit tests for the new method.

Related: https://github.com/canonical/hardware-observer-operator/issues/61

### DIFF
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -32,10 +32,10 @@ class RedfishHelper:
         """
 
         @ttl_cache(ttl=ttl)
-        def _discover(host: str, username: str, password: str) -> bool:
+        def _discover(host: str) -> bool:
             """Return True if redfish service has been discovered."""
             try:
-                redfish_obj = redfish_client(base_url=host, username=username, password=password)
+                redfish_obj = redfish_client(base_url=host, username="", password="")
                 redfish_obj.login(auth="session")
             except RetriesExhaustedError:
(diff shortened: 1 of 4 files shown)


prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -26,15 +32,27 @@ def get_cached_discover_method(ttl: int) -> Callable:
...
-        def _discover() -> bool:
-            """Return true if redfish services have been discovered."""
-            logger.info("Discovering redfish services...")
-            services = redfish.discover_ssdp()
-            if len(services) == 0:
-                logger.info("No redfish services discovered")
-                return False
-            logger.debug("Discovered redfish services: %s", services)
-            return True
+        def _discover(host: str, username: str, password: str) -> bool:
```
Pjack, line 35: Based on your code here. we don't need username/password.
https://github.com/canonical/hardware-observer-operator/pull/69/files

dashmage: You're right, I missed that. Thanks for pointing it out. I've made the required changes in the latest commit.

sudeephb (approved):
LGTM

Pjack (approved):
LGTM

PR #39, opened 2023-10-06
Adopt juju 3.1

dashmage:


.github/workflows/check.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+  func:
+    uses: canonical/bootstack-actions/.github/workflows/func.yaml@v2
+    needs: lint-unit
+    with:
+      command: make functional
+      juju-channel: "3.1/stable"
+      nested-containers: true
+      python-version: "3.10"
+      timeout-minutes: 120
```
rgildein, line 31: not-blocking: I think 60 min will be enough here

rgildein (approved):
LGTM

.github/workflows/check.yaml
dashmage: Done!

rgildein (approved):

.github/workflows/check.yaml
```diff
@@ -0,0 +1,32 @@
+name: Check workflow running linter, unit and functional tests
```
agileshaw, line 1: We also need to make changes in `sonar.yaml` ([L#9](https://github.com/canonical/prometheus-hardware-exporter/blob/main/.github/workflows/sonar.yaml#L9) and[ L#17](https://github.com/canonical/prometheus-hardware-exporter/blob/main/.github/workflows/sonar.yaml#L17)) to track this new workflow name.

agileshaw (approved):
LGTM, thanks

PR #45, opened 2023-11-15
feat: Add metrics ipmi_dcmi_power_consumption_rate

jneo8:
Add new ipmi-dcmi metrics
Rate = power consumption / maximum power consumption

proposal to fix: https://github.com/canonical/hardware-observer-operator/issues/97

Should merge together with: https://github.com/canonical/hardware-observer-operator/pull/116


prometheus_hardware_exporter/collectors/lshw.py
```diff
@@ -0,0 +1,32 @@
...
+
+class LSHW(Command):
+    """Command line tool for lshw."""
+
+    prefix = ""
+    command = "lshw"
+
+    def get_powers(self) -> List[Dict[str, str]]:
+        """Get power class data."""
+        result = self("-c power -json")
```
aieri, line 19: why did you choose to use lshw instead of dmidecode?
On a server I am testing I get similar but not identical results:
```
$ sudo dmidecode -t 39 | grep 'Power Capacity'
	Max Power Capacity: 550 W
	Max Power Capacity: 550 W

$ sudo lshw -c power | grep capacity
       capacity: 550mWh
       capacity: 550mWh
```

(obviously Wh != W)

prometheus_hardware_exporter/collectors/lshw.py
```diff
@@ -0,0 +1,32 @@
...
+import json
+from logging import getLogger
+from typing import Dict, List
+
+from ..utils import Command
+
+logger = getLogger(__name__)
+
+
+class LSHW(Command):
```
aieri, line 11: should we cache results so we only subprocess out once? PSU ratings won't change over the lifetime of a server

prometheus_hardware_exporter/collector.py
```diff
@@ -349,11 +358,39 @@ def fetch(self) -> List[Payload]:
...
+        # Suppose redundancy enable make denominator smaller
+        # and alert is more easy to fire.
+        if not get_ps_redundancy_ok:
+            ps_redundancy = True
+
+        power_capacities = self.lshw.get_power_capacities()
+        maximum_power_capacity = (
+            sum(power_capacities) / len(power_capacities)
+            if ps_redundancy
+            else sum(power_capacities)
```
aieri, line 372: I don't recall ever seeing a server using two power supplies in parallel or in a combined fashion. Is it even possible?

aieri (commented):
Thank you, this solves the impossibility of a one size fits all approach (including inconsistent power budgets across the hardware-observer application). I have left a few comments

prometheus_hardware_exporter/collectors/lshw.py
jneo8: $$P(W) = E(Wh) / t(h)$$

Yes, using `W` will be more correct. Let's change to dmidecode.

prometheus_hardware_exporter/collector.py
jneo8: It's possible according to the hardware design I know, it's rare case.
If you think we don't need to support this I can remove it.

jneo8: Another issue is if there are two powers with different power capacities, We don't have a way to know which power is in used from the command. Do you think average make sense here?

prometheus_hardware_exporter/collectors/lshw.py
jneo8: Updated

prometheus_hardware_exporter/collector.py
```diff
@@ -339,6 +342,12 @@ def specifications(self) -> List[Specification]:
                 documentation="Indicates if the ipmi dcmi command is successful or not",
                 metric_class=GaugeMetricFamily,
             ),
+            Specification(
+                name="ipmi_dcmi_power_consumption_rate",
+                documentation="Current power capacity rate",
```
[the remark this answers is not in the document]
jneo8: Updated

prometheus_hardware_exporter/collector.py
```diff
@@ -339,6 +342,12 @@ def specifications(self) -> List[Specification]:
                 documentation="Indicates if the ipmi dcmi command is successful or not",
                 metric_class=GaugeMetricFamily,
             ),
+            Specification(
+                name="ipmi_dcmi_power_consumption_rate",
```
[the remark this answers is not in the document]
jneo8: Updated

prometheus_hardware_exporter/collectors/lshw.py
jneo8: Add cache for dmidecode.

prometheus_hardware_exporter/collector.py
```diff
@@ -349,11 +360,39 @@ def fetch(self) -> List[Payload]:
...
 
+        get_ps_redundancy_ok, ps_redundancy = self.ipmi_tool.get_ps_redundancy()
+        # Because we fail to get the redundancy config from the server,
+        # Suppose redundancy enable make denominator smaller
+        # and alert is more easy to fire.
+        if not get_ps_redundancy_ok:
+            ps_redundancy = True
+
+        power_capacities = self.dmidecode.get_power_capacities()
+        maximum_power_capacity = (
```
Pjack, line 371: Please add some comment to explain the fomula. It's different when ps_redundancy is true/false.

tests/unit/test_collector.py
```diff
@@ -303,10 +306,34 @@ def test_31_ipmi_dcmi_collector_installed_and_okay(self):
...
+        """Test ipmi dcmi collector can fetch correct number of metrics."""
+        ipmi_dcmi_collector = IpmiDcmiCollector(Mock())
+        ipmi_dcmi_collector.ipmi_dcmi = Mock()
+        ipmi_dcmi_collector.ipmi_tool = Mock()
+        ipmi_dcmi_collector.dmidecode = Mock()
 
         mock_dcmi_payload = {"current_power": 105}
 
         ipmi_dcmi_collector.ipmi_dcmi.get_current_power.return_value = mock_dcmi_payload
+        ipmi_dcmi_collector.ipmi_tool.get_ps_redundancy.return_value = (False, True)
```
Pjack, line 335: It's okay to test with (False, True), However, if I peek into the function, it's impoosible to return (False, True).


Furthermore, do we want to test different combination to verify the fomula ? 
1. ps_redundancy  =   (True, True)  ,  (True, False) ,  (False, False)
2. get_power_capacities = [] ,  [1000], [1000, 1000] , [1000, 1500]

expected value of maximum_power_capacity and power_capacity_percentage

tests/unit/test_ipmi_dcmi.py
```diff
@@ -31,3 +32,22 @@ def test_01_get_current_power_parse_failure(self, mock_call):
...
+    """Test the IpmiTool class."""
+
+    @patch.object(Command, "__call__")
+    def test_00_get_ps_redundancy_success(self, mock_call):
+        with open(IPMITOOL_SDR_PS_SAMPLE_OUTPUT, "r") as content:
+            mock_call.return_value = Result(content.read(), None)
+        ipmitool = IpmiTool()
+        ps_redundancy = ipmitool.get_ps_redundancy()
+        self.assertEqual(ps_redundancy, (True, True))
+
```
Pjack, line 47: Need a case to return (True, False)

Pjack (changes requested):

prometheus_hardware_exporter/collectors/ipmi_dcmi.py
```diff
@@ -11,6 +11,32 @@
...
+
+    prefix = ""
+    command = "ipmitool"
+
+    def get_ps_redundancy(self) -> Tuple[bool, bool]:
+        """Get power supply redundancy.
+
+        returns:
+            - ok - Get redundancy success?
+            - redundancy - enable redundancy?
```
sudeephb, line 25: This doesn't look very clear. Can you change the description to something like - 
`ok - True if fetching redundancy info is successful` 
`redundancy - True if redundancy is enabled`?

Pjack (approved):

PR #49, opened 2023-11-17
fix(redfish): Make redfish storage name in URI dynamic.

dashmage:
The redfish storage name in the URI was initially hardcoded to "Storage". On some servers which did not conform to the schema specification, this name was provided differently, eg: Storages.

This change aims to find the storage uri name dynamically while fetching the storage controller and storage drive data.

Fixes https://github.com/canonical/hardware-observer-operator/issues/91, https://github.com/canonical/hardware-observer-operator/issues/108

### DIFF
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -65,6 +65,7 @@ class RedfishHelper:
         self.timeout = config.redfish_client_timeout
         self.max_retry = config.redfish_client_max_retry
         self.redfish_obj: HttpClient
+        self.systems_root_uri: str = "/redfish/v1/Systems/"
 
     def __enter__(self) -> Self:
         """Login to redfish while entering context manager."""
@@ -208,6 +209,18 @@ class RedfishHelper:
         logger.debug("Processor data: %s", processor_data)
         return processor_count, processor_data
 
+    def _storage_root_uri(self, system_id: str, storage_name: str) -> str:
+        """Return formatted URI string for the root storage location.
+
+        storage_root_uri is in the format of "systems_root_uri/system_id/storage_name/"
+
+        eg: For "/redfish/v1/Systems/S1/Storage/",
+          systems_root_uri: "/redfish/v1/Systems"
+          system_id: "S1"
+          storage_name" "Storage"
+        """
+        return self.systems_root_uri + f"{system_id}/{storage_name}/"
+
     def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
         """Return storage controller data and count.
 
@@ -241,11 +254,7 @@ class RedfishHelper:
         storage_controller_data: Dict[str, List] = {}
         # storage controllers on each system
         storage_controller_count: Dict[str, int] = {}
-        systems_root_uri = "/redfish/v1/Systems/"
 
-        # storage_root_uri becomes systems_root_uri/system_id/storage_name
-        # eg: "/redfish/v1/Systems/S1/Storage"
-        storage_root_uri_pattern = systems_root_uri + "{}/{}"
         logger.info("Getting storage controller data...")
 
         try:
@@ -257,23 +266,21 @@ class RedfishHelper:
         for system_id in system_ids:
             storage_controller_count[system_id] = 0
 
-            # finding storage name used in URI
+            # finding storage name used in URI by querying Systems dictionary
             # eg: if storage root URI is /redfish/v1/Systems/S1/Storages
-            # then storage_name becomes "Storages"
-            system_dict = self.redfish_obj.get(systems_root_uri + "/" + system_id).dict
+            # then storage_name is "Storages"
+            system_dict = self.redfish_obj.get(self.systems_root_uri + system_id).dict
             storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]
 
             # List of storage ids
             storage_ids: List[str] = redfish_utilities.collections.get_collection_ids(
-                self.redfish_obj, storage_root_uri_pattern.format(system_id, storage_name)
+                self.redfish_obj, self._storage_root_uri(system_id, storage_name)
             )
 
             storage_controller_data_in_curr_system = []
             for storage_id in storage_ids:
                 # eg: /redfish/v1/Systems/1/Storage/XYZ123
-                curr_storage_uri = (
-                    storage_root_uri_pattern.format(system_id, storage_name) + "/" + storage_id
-                )
+                curr_storage_uri = self._storage_root_uri(system_id, storage_name) + storage_id
 
                 # list of storage controllers for that storage id
                 storage_controllers_list: List[Dict] = self.redfish_obj.get(curr_storage_uri).dict[
@@ -411,11 +418,6 @@ class RedfishHelper:
         storage_drive_data: Dict[str, List] = {}
         # storage drives on each system
         storage_drive_count: Dict[str, int] = {}
-        systems_root_uri = "/redfish/v1/Systems/"
-
-        # storage_root_uri becomes systems_root_uri/system_id/storage_name
-        # eg: "/redfish/v1/Systems/S1/Storage"
-        storage_root_uri_pattern = systems_root_uri + "{}/{}"
         logger.info("Getting storage drive data...")
 
         try:
@@ -427,22 +429,20 @@ class RedfishHelper:
         for system_id in system_ids:
             storage_drive_count[system_id] = 0
 
-            # finding storage name used in URI
+            # finding storage name used in URI by querying Systems dictionary
             # eg: if storage root URI is /redfish/v1/Systems/S1/Storages
-            # then storage_name becomes "Storages"
-            system_dict = self.redfish_obj.get(systems_root_uri + "/" + system_id).dict
+            # then storage_name is "Storages"
+            system_dict = self.redfish_obj.get(self.systems_root_uri + system_id).dict
             storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]
 
             storage_ids: List[str] = redfish_utilities.collections.get_collection_ids(
-                self.redfish_obj, storage_root_uri_pattern.format(system_id, storage_name)
+                self.redfish_obj, self._storage_root_uri(system_id, storage_name)
             )
 
             storage_drive_data_in_curr_system: List[Dict] = []
             for storage_id in storage_ids:
                 # eg: /redfish/v1/Systems/1/Storage/XYZ123/
-                curr_storage_uri = (
-                    storage_root_uri_pattern.format(system_id, storage_name) + "/" + storage_id
-                )
+                curr_storage_uri = self._storage_root_uri(system_id, storage_name) + storage_id
 
                 # list of storage drives for that storage id
                 storage_drives_list: List[Dict] = self.redfish_obj.get(curr_storage_uri).dict[
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -266,9 +266,11 @@ class RedfishHelper:
         for system_id in system_ids:
             storage_controller_count[system_id] = 0
 
-            # finding storage name used in URI by querying Systems dictionary
+            # finding storage name used in URI by querying resource dictionary of system
             # eg: if storage root URI is /redfish/v1/Systems/S1/Storages
             # then storage_name is "Storages"
+            # The storage name in URI can sometimes differ from the standard "Storage"
+            # so we need to query it dynamically to not break the API calls.
             system_dict = self.redfish_obj.get(self.systems_root_uri + system_id).dict
             storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]
 
@@ -429,9 +431,11 @@ class RedfishHelper:
         for system_id in system_ids:
             storage_drive_count[system_id] = 0
 
-            # finding storage name used in URI by querying Systems dictionary
+            # finding storage name used in URI by querying resource dictionary of system
             # eg: if storage root URI is /redfish/v1/Systems/S1/Storages
             # then storage_name is "Storages"
+            # The storage name in URI can sometimes differ from the standard "Storage"
+            # so we need to query it dynamically to not break the API calls.
             system_dict = self.redfish_obj.get(self.systems_root_uri + system_id).dict
             storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]
 
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -25,6 +25,8 @@ logger = getLogger(__name__)
 class RedfishHelper:
     """Helper function for redfish."""
 
+    systems_root_uri: str = "/redfish/v1/Systems/"
+
     @staticmethod
     def get_cached_discover_method(ttl: int) -> Callable:
         """Return the cached discover method.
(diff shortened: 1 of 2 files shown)


dashmage:
There's duplicate code between the `get_storage_drive_data` and the `get_storage_controller_data` methods while navigating the URIs and fetching the storage related data. Currently trying to refactor the common code into a separate method.

tests/unit/test_redfish.py
```diff
@@ -555,11 +558,92 @@ def mock_get_response(uri):
...
         mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s1/Storage/STOR2")
 
+    @patch("prometheus_hardware_exporter.collectors.redfish.redfish_client")
+    @patch(
+        "prometheus_hardware_exporter.collectors.redfish.redfish_utilities.collections.get_collection_ids"  # noqa: E501
+    )
+    @patch(
+        "prometheus_hardware_exporter.collectors.redfish.redfish_utilities.systems.get_system_ids"
+    )
+    def test_12_non_standard_storage_uri_name(
```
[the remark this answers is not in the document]
Pjack: The logic in test_11 and test_12 are almost the same. Please think about how to reduce it.

Pjack (changes requested):

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -252,14 +257,24 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
         for system_id in system_ids:
             storage_controller_count[system_id] = 0
 
+            # finding storage name used in URI
+            # eg: if storage_root_uri is /redfish/v1/Systems/S1/Storages
+            # then storage_name becomes "Storages"
+            system_dict = self.redfish_obj.get(systems_root_uri + "/" + system_id).dict
+            storage_root_uri = system_dict["Storage"]["@odata.id"]
+            storage_name = storage_root_uri.split("/")[-1]
```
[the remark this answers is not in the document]
dashmage: Someone reading `storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]`  might not understand how we can find the storage name from the complex expression. So I decided to use another variable for the `storage_root_uri` which makes this more clearer.

First find `storage_root_uri` --> `/redfish/v1/Systems/S1/Storage`
Now split the last part from `storage_root_uri` for getting `Storage`. 

If you still find `storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]` to be better, I can make the change.

tests/unit/test_redfish.py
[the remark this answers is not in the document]
dashmage: > I'm curious why we must use "numbers" to sort the test cases in code.
Whenever you add new test case in the middele , it impacts all the other test cases.

I'm onboard with removing these test numbers altogether. It definitely is a pain to re-order them while adding a new test case in the middle.

> The logic in test_11 and test_12 are almost the same. Please think about how to reduce it.

This is because the testing scenario for both is only slightly different -- storage name in URI is either `Storage` or `Storages`. I can put this as part of one test but then it won't be testing only one thing. Let me think about if there's a way to simplify these tests.

prometheus_hardware_exporter/collectors/redfish.py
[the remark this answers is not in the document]
Pjack: there is no magic of this syntax   (read a value in a dictionary and then split by "/"). It is straightforward to me.
`system_dict["Storage"]["@odata.id"].split("/")[-1]`
My understanding between two syntaxs are almost the same. Therefore, if you worry about other people cannot understand what you want to parse. it's better to give some example in the comment.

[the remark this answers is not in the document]
dashmage: Alright in that case, I'll go ahead with your suggestion and leave a helpful comment.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -409,13 +428,24 @@ def get_storage_drive_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
         for system_id in system_ids:
             storage_drive_count[system_id] = 0
 
+            # finding storage name used in URI
+            # eg: if storage_root_uri is /redfish/v1/Systems/S1/Storages
+            # then storage_name becomes "Storages"
+            system_dict = self.redfish_obj.get(systems_root_uri + "/" + system_id).dict
+            storage_root_uri = system_dict["Storage"]["@odata.id"]
+            storage_name = storage_root_uri.split("/")[-1]
```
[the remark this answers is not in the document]
dashmage: Made the required changes here as well.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -240,7 +241,11 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
         storage_controller_data: Dict[str, List] = {}
         # storage controllers on each system
         storage_controller_count: Dict[str, int] = {}
-        storage_root_uri_pattern = "/redfish/v1/Systems/{}/Storage"
+        systems_root_uri = "/redfish/v1/Systems/"
+
+        # storage_root_uri becomes systems_root_uri/system_id/storage_name
+        # eg: "/redfish/v1/Systems/S1/Storage"
+        storage_root_uri_pattern = systems_root_uri + "{}/{}"
```
rgildein, line 248: Would not be better to use pathlib instead of predefined string? Or using simple function to generate url?
e.g.:
```python
def _storage_uri(self, system_id, storage_name):
    return self.systems_root_uri + f"{system_id}/{storage_name}"
```

tests/unit/test_redfish.py
```diff
@@ -555,11 +565,69 @@ def mock_get_response(uri):
...
+            # response for GET request to /redfish/v1/Systems/<sys_id>/
+            elif "Systems" in uri:
+                response.dict = {"Storage": {"@odata.id": "/redfish/v1/Systems/sX/Storages"}}
+            return response
+
+        mock_redfish_obj.get.side_effect = mock_get_response
+
+        # storage controller
+        with RedfishHelper(Mock()) as helper:
+            _sc_count, _sc_data = helper.get_storage_controller_data()
```
Pjack, line 615: Why we didn't check the expected value of sc_count and sc_data? 

The testing logic between test_non_standard_storage_uri_name and test_get_storage_controller_data_success should be the same. We only need to simulate different `mock_get_response`.

tests/unit/test_redfish.py
```diff
@@ -555,11 +565,69 @@ def mock_get_response(uri):
...
+        mock_get_system_ids.assert_called_once_with(mock_redfish_obj)
+        # test if "Storages" is present in redfish GET request URI
+        mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s1/Storages/STOR1")
+
+        # storage drives
+        with RedfishHelper(Mock()) as helper:
+            _drive_count, _drive_data = helper.get_storage_drive_data()
```
Pjack, line 622: ditto, why we do not check the expected result of drive_count and drive_data?

tests/unit/test_redfish.py
```diff
@@ -555,11 +565,69 @@ def mock_get_response(uri):
         mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s1/Storage/STOR1")
         mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s1/Storage/STOR2")
```
Pjack, line 567: Although these code are not written by you, I think it's unnecessary to check the API call url. Checking the expected output is good enough to me.

tests/unit/test_redfish.py
```diff
@@ -555,11 +565,69 @@ def mock_get_response(uri):
...
+
+        mock_redfish_obj.get.side_effect = mock_get_response
+
+        # storage controller
+        with RedfishHelper(Mock()) as helper:
+            _sc_count, _sc_data = helper.get_storage_controller_data()
+        mock_get_system_ids.assert_called_once_with(mock_redfish_obj)
+        # test if "Storages" is present in redfish GET request URI
+        mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s1/Storages/STOR1")
+
```
Pjack, line 619: I prefer to check the expected output rather than checking the url.

tests/unit/test_redfish.py
```diff
@@ -555,11 +565,69 @@ def mock_get_response(uri):
...
+        # storage drives
+        with RedfishHelper(Mock()) as helper:
+            _drive_count, _drive_data = helper.get_storage_drive_data()
+        # test if "Storages" is present in redfish GET request URI
+        mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s1/Storages/STOR1/Drives/d11")
```
Pjack, line 624: I prefer to check the expected output rather than checking the url.

Pjack (changes requested):

tests/unit/test_redfish.py
dashmage: > Why we didn't check the expected value of sc_count and sc_data?
> The testing logic between test_non_standard_storage_uri_name and test_get_storage_controller_data_success should be the same. We only need to simulate different mock_get_response.

I was doing this to not duplicate the code being tested with `test_get_storage_controller_data_success`. The purpose of `test_non_standard_storage_uri_name` was only to check in case there's a non standard storage name being used, we were actually querying the right endpoint for the redfish API. The actual storage controller data or count doesn't matter because if we're sure about the endpoint, then we're testing the ability to obtain the data, count from `test_get_storage_controller_data_success`.

tests/unit/test_redfish.py
dashmage: IMO for `test_get_storage_controller_data_success`, we should be testing for the expected output and the correct redfish API call. It would be nice to test that we're calling the right endpoints with `redfish_obj.get` in order to fetch the correct results.

tests/unit/test_redfish.py
```diff
@@ -481,14 +479,23 @@ def mock_get_response(uri):
         mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s2/Processors")
         mock_redfish_obj.get.assert_any_call("/redfish/v1/Systems/s2/Processors/p21")
 
+    @patch("prometheus_hardware_exporter.collectors.redfish.redfish_client")
+    def test__storage_root_uri(self, mock_redfish_client):
```
Pjack, line 483: We should test different pattern with different expected_uri. (standard and non-standard.)

tests/unit/test_redfish.py
dashmage: Alright fair enough, I understand your point about the method being a black box and not having to test any implementation details. In that case we would need to modify almost all the test methods to remove the assertions for the API calls since the same point would apply to all of them and we will need to be consistent.

dashmage: In the same vein, we do not need to include the assertion below as well since we don't need to worry about whether the `redfish_utilities.systems.get_system_ids` method is called with `redfish_obj`. We could possibly obtain the system ids in other ways.

```python
mock_get_system_ids.assert_called_once_with(mock_redfish_obj)
```

I'll go ahead and remove this from the test methods as well.

tests/unit/test_redfish.py
dashmage: Changed in the latest patch.

tests/unit/test_redfish.py
dashmage: Changed in the latest patch.

Pjack (approved):
LGTM

tests/unit/test_redfish.py
chanchiwai-ray: Yes, I agree with @Pjack, testing the end product of that function is necessary. However, if you do want to test if you can get the correct storage name, then you might want to make the logic [here](https://github.com/canonical/prometheus-hardware-exporter/blob/6b1ce0880714a2a6e672145d11e2cc7bcd554bdb/prometheus_hardware_exporter/collectors/redfish.py#L273) as a separate function, and test it.

chanchiwai-ray: But that being said, I think you don't need to remove that assertion, but add a check about the end result is fine.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -252,14 +266,21 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
         for system_id in system_ids:
             storage_controller_count[system_id] = 0
 
+            # finding storage name used in URI by querying Systems dictionary
```
jneo8, line 269: Can you explain in the comment why we want to find storage name instead of hard code?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -252,14 +266,21 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
...
+            # eg: if storage root URI is /redfish/v1/Systems/S1/Storages
+            # then storage_name is "Storages"
+            system_dict = self.redfish_obj.get(self.systems_root_uri + system_id).dict
```
jneo8, line 272: system_dict is a confused variable name. Maybe should relate to the essence of the return object's meaning.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -252,14 +266,21 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
...
+            storage_name = system_dict["Storage"]["@odata.id"].split("/")[-1]
```
jneo8, line 273: Can here raise the KeyError if `Storage` is not in the dict? If not please ignore this comment.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -64,6 +65,7 @@ def __init__(self, config: Config) -> None:
         self.timeout = config.redfish_client_timeout
         self.max_retry = config.redfish_client_max_retry
         self.redfish_obj: HttpClient
+        self.systems_root_uri: str = "/redfish/v1/Systems/"
```
jneo8, line 68: It's more like a static variable.

jneo8 (changes requested):
Just some nitpick of naming and comment.

tests/unit/test_redfish.py
jneo8: The main purpose of this testing is to make sure our call to redfish can compatible with different storage name.

Just want to point out url checking is the major checking item here.

The return object is not what we care about here, although you can say it indirect prove it's working and adding one more checking here is not a big deal

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Good point, I'll mention about how the name can sometimes be non-standard and we need to query it dynamically.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: As part of the redfish schema specification, the dictionary of resources from the systems API call should have a `Storage` resource that points to the URI for fetching the storage data. Unless the system doesn't have any storage at all which I highly unlikely I believe.

But it's a good thing to think about.

prometheus_hardware_exporter/collectors/redfish.py
dashmage: The return object is essentially a dictionary of the resources returned by querying the systems endpoint (eg: `/redfish/v1/Systems/S1`). 

Maybe a better name could be `system_resources_dict`?

prometheus_hardware_exporter/collectors/redfish.py
dashmage: Making the change in the latest patch

prometheus_hardware_exporter/collectors/redfish.py
jneo8: `system_resources` is better.

sudeephb (approved):
LGTM

jneo8 (approved):

PR #52, opened 2023-11-21
fix: General err handling for collector

jneo8:
The error handler will try to output the failed metrics and make sure the single collector's bug won't affect other collectors.

---

Review together with: https://github.com/canonical/hardware-observer-operator/pull/117

### DIFF
--- a/prometheus_hardware_exporter/core.py
+++ b/prometheus_hardware_exporter/core.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass, field
 from logging import getLogger
 from typing import Any, Dict, Iterable, List, Type
 
-from prometheus_client.metrics_core import Metric
+from prometheus_client.metrics_core import GaugeMetricFamily, Metric
 from prometheus_client.registry import Collector
 
 from .config import Config
@@ -89,6 +89,22 @@ class BlockingCollector(Collector):
             A list of specification.
         """
 
+    @property
+    def failed_metrics(self) -> Iterable[Metric]:
+        """Defines the metrics return when collect func fail.
+
+        Yields:
+            metrics: the internal metrics
+        """
+        name = self.__class__.__name__
+        metric = GaugeMetricFamily(
+            name=f"{name.lower()}_collector_failed",
+            labels=[],
+            documentation=f"{name} Collector fail to fetch metrics",
+        )
+        metric.add_metric(labels=[], value=1)
+        yield metric
+
     def init_default_datastore(self, payloads: List[Payload]) -> None:
         """Initialize or fill data the store with default values.
 
@@ -101,7 +117,7 @@ class BlockingCollector(Collector):
                     name=payload.name, labels=payload.labels, value=0.0
                 )
 
-    def collect(self) -> Iterable[Metric]:
+    def collect(self) -> Iterable[Metric]:  # pylint: disable=R1710
         """Fetch data and update the internal metrics.
 
         This is a callback method that is used internally within
@@ -111,21 +127,28 @@ class BlockingCollector(Collector):
         Yields:
             metrics: the internal metrics
         """
-        payloads = self.fetch()
-        self.init_default_datastore(payloads)
-        processed_payloads = self.process(payloads, self._datastore)
+        # The general exception hanlder will try to make sure the single
+        # collector's bug will only change the metrics output to failed_metrics
+        # and also make sure other collectors are still working.
+        try:
+            payloads = self.fetch()
+            self.init_default_datastore(payloads)
+            processed_payloads = self.process(payloads, self._datastore)
 
-        # unpacked and create metrics
-        for payload in processed_payloads:
-            spec = self._specs[payload.name]
-            # We have to ignore the type checking here, since the subclass of
-            # any metric family from prometheus client adds new attributes and
-            # methods.
-            metric = spec.metric_class(  # type: ignore[call-arg]
-                name=spec.name, labels=spec.labels, documentation=spec.documentation
-            )
-            metric.add_metric(  # type: ignore[attr-defined]
-                labels=payload.labels, value=payload.value
-            )
-            yield metric
-            self._datastore[payload.uuid] = payload
+            # unpacked and create metrics
+            for payload in processed_payloads:
+                spec = self._specs[payload.name]
+                # We have to ignore the type checking here, since the subclass of
+                # any metric family from prometheus client adds new attributes and
+                # methods.
+                metric = spec.metric_class(  # type: ignore[call-arg]
+                    name=spec.name, labels=spec.labels, documentation=spec.documentation
+                )
+                metric.add_metric(  # type: ignore[attr-defined]
+                    labels=payload.labels, value=payload.value
+                )
+                yield metric
+                self._datastore[payload.uuid] = payload
+        except Exception as e:  # pylint: disable=W0718
+            logger.error(e)
+            yield from self.failed_metrics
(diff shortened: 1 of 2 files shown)


chanchiwai-ray:
This changes might potentially affect all collectors, can you add functional test? i.e. enable a collector, mock the fetch function to raise exception, and see if it outputs the failed metrics? 

Or at least test it manually

Pjack:
Please also update

- [ ] test plan (mindmap): https://drive.google.com/file/d/1sa2TD78WtAUzEUzbQMvcJwISMjKjTqEp/view?usp=drive_link

- [ ] test case: https://docs.google.com/spreadsheets/d/1KiQ5cS36zfjf8V32RoD1RKx9q01AhgoKvBQRkonQv5o/edit#gid=498657749

jneo8:
> Ideally, the collector itself should be responsible to catch general error, and produce a payload when it is failed.
> This changes might potentially affect all collectors, can you add functional test? i.e. enable a collector, mock the fetch function to raise exception, and see if it outputs the failed metrics?
> Or at least test it manually

Yes, unit test and manually test are done.

Agree with your point that collector should responsible to catch it's error. But won't it be the same if we need to have this practice on all the collectors?
Instead of catch general error on each, reduce the possible to make human error, like every time you create a new collector and you need to remember this practice,  adding catch in `core` module maybe make more effectually and safe from my point of view.

Pjack (changes requested):
Request to update the test plan and test case.

Pjack (approved):

prometheus_hardware_exporter/core.py
```diff
@@ -101,7 +117,7 @@ def init_default_datastore(self, payloads: List[Payload]) -> None:
                     name=payload.name, labels=payload.labels, value=0.0
                 )
 
-    def collect(self) -> Iterable[Metric]:
+    def collect(self) -> Iterable[Metric]:  # pylint: disable=R1710
```
rgildein, line 120: I do not think that disabling inconsistent return statements is right solution. You should provide `Iterable[Metric] | ...`.

prometheus_hardware_exporter/core.py
```diff
@@ -89,6 +89,22 @@ def specifications(self) -> List[Specification]:
...
+        Yields:
+            metrics: the internal metrics
+        """
+        name = self.__class__.__name__
+        metric = GaugeMetricFamily(
+            name=f"{name.lower()}_collector_failed",
+            labels=[],
+            documentation=f"{name} Collector fail to fetch metrics",
+        )
+        metric.add_metric(labels=[], value=1)
```
rgildein, line 105:
```suggestion
        metric = GaugeMetricFamily(
            name=f"{name.lower()}_collector_failed",
            documentation=f"{name} Collector fail to fetch metrics",
            value=1,
            labels=[],   
        )
```

prometheus_hardware_exporter/core.py
```diff
@@ -111,21 +127,28 @@ def collect(self) -> Iterable[Metric]:
...
+                    name=spec.name, labels=spec.labels, documentation=spec.documentation
+                )
+                metric.add_metric(  # type: ignore[attr-defined]
+                    labels=payload.labels, value=payload.value
+                )
+                yield metric
+                self._datastore[payload.uuid] = payload
+        except Exception as e:  # pylint: disable=W0718
+            logger.error(e)
+            yield from self.failed_metrics
```
rgildein, line 154: Would it not be better if failed metrics contains some information about why collection failed? Maybe via label?

jneo8: It's for unexpected error. So it's weird to have a label of it.

prometheus_hardware_exporter/core.py
jneo8: Nice one, I don't know we can even do this.

jneo8: Updated

prometheus_hardware_exporter/core.py
jneo8: Updated

prometheus_hardware_exporter/core.py
rgildein: It will mean that we will have more exception not only the generic one, if it's possible. e.g.
```python
except ExceptionA as exc:
    logger.exception(exc)
    yield from self.failed_metrics(label="ExceptionA")
except ExceptionB as exc:
    logger.exception(exc)
    yield from self.failed_metrics(label="ExceptionB")
except Exception as exc:
    logger.exception(exc)
    yield from self.failed_metrics(label="Unknown")
```

jneo8: If that's the case, it should be handle inside each controller from my view.

prometheus_hardware_exporter/core.py
```diff
@@ -89,6 +89,25 @@ def specifications(self) -> List[Specification]:
             A list of specification.
         """
 
+    @property
+    def failed_metrics(self) -> Iterable[Metric]:
+        """Defines the metrics return when collect func fail.
```
sudeephb, line 94:
```suggestion
        """Defines the metrics to be returned when a collector function fails.
```

prometheus_hardware_exporter/core.py
```diff
@@ -89,6 +89,25 @@ def specifications(self) -> List[Specification]:
...
+
+        Yields:
+            metrics: the internal metrics
+        """
+        name = self.__class__.__name__
+        metric = GaugeMetricFamily(
+            name=f"{name.lower()}_collector_failed",
+            documentation=f"{name} Collector fail to fetch metrics",
```
dashmage, line 102: typo: Collector *failed* to fetch metrics

tests/unit/test_collector.py
```diff
@@ -1235,3 +1235,31 @@ def test_210_redfish_create_smart_storage_health_metric_payload(self):
...
+            ),
+            (
+                IpmiSensorsCollector,
+                "ipmisensorscollector_collector_failed",
+                {"collector": "IpmiSensorsCollector"},
+            ),
+        ]:
+            collector = collector_cls(Mock())
+            collector.fetch = Mock()
+            collector.fetch.side_effect = Exception("Unknow error")
```
dashmage, line 1259: typo: *Unknown* error

dashmage (approved):
A few typos to fix. Otherwise LGTM :)

prometheus_hardware_exporter/core.py
jneo8: Updated

tests/unit/test_collector.py
jneo8: Updated

prometheus_hardware_exporter/core.py
jneo8: Updated

prometheus_hardware_exporter/core.py
rgildein: I'm not an expert here, but I do not understand what is controller here and how it could handle those exceptions.

jneo8: The purpose is that we try to catch those unexpected error and try not to make exporter whole broken. It's kind of defensive purpose.

So if we know what exactly the error is, we should handle that inside collector.

rgildein: Ok, that make sense to me. Thanks

rgildein: And I think you meant collector by "each controller from my view."  as I understand it from https://github.com/canonical/hardware-observer-operator/pull/116.

rgildein (approved):
LGTM thanks

sudeephb (approved):

jneo8:
Waiting for https://github.com/canonical/hardware-observer-operator/pull/117

PR #58, opened 2023-12-05
handle different storage controller responses

sdemircan:
While checking storage controller data in get_storage_controller_data function API response varies between different ILO versions and different vendors(ie. Inspur).

### DIFF
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -223,6 +223,28 @@ class RedfishHelper:
         """
         return cls.systems_root_uri + f"{system_id}/{storage_name}"
 
+    def _find_storage_controller_data(self, storage_controllers: Dict) -> List[Dict]:
+        """Return storage controller list.
+
+        Returns:
+            storage_controller_list: list of storage controllers for 
+            requested storage id.
+
+        """
+        if "StorageControllers" in storage_controllers:
+            storage_controllers_list: List[Dict] = storage_controllers.get("StorageControllers")
+        elif "Controllers" in storage_controllers:
+            storage_controllers_list: List[Dict] = []
+            controller_list = storage_controllers.get("Controllers")
+            for controller in controller_list.values():
+                all_controller_data = self.redfish_obj.get(controller).dict
+                for controller in all_controller_data["Members"]:
+                    storage_controllers_list.append(self.redfish_obj.get(controller["@odata.id"]).dict)
+        else:
+            raise KeyError("Couldn't fetch storage controller data")
+        
+        return storage_controllers_list
+
     def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
         """Return storage controller data and count.
 
@@ -291,20 +313,14 @@ class RedfishHelper:
                 )
 
                 # list of storage controllers for that storage id
-                storage_controllers = self.redfish_obj.get(curr_storage_uri).dict
-                if "StorageControllers" in storage_controllers:
-                    storage_controllers_list: List[Dict] = storage_controllers.get("StorageControllers")
-                if "Controllers" in storage_controllers:
-                    storage_controllers_list: List[Dict] = []
-                    controller_list = storage_controllers.get("Controllers")
-                    for controller in controller_list.values():
-                        all_controller_data = self.redfish_obj.get(controller).dict
-                        for controller in all_controller_data["Members"]:
-                            storage_controllers_list.append(self.redfish_obj.get(controller["@odata.id"]).dict)
-
+                storage_controllers_list: List[Dict] = self._find_storage_controller_data(
+                    self.redfish_obj.get(curr_storage_uri).dict
+                )
+                storage_controller_count[system_id] += len(storage_controllers_list)
 
                 # picking out the required data from each storage controller in the list
                 for data in storage_controllers_list:
+                    print(data)
                     storage_controller_data_in_curr_system.append(
                         {
                             "storage_id": storage_id,
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -221,7 +221,7 @@ class RedfishHelper:
           system_id: "S1"
           storage_name" "Storage"
         """
-        return cls.systems_root_uri + f"{system_id}/{storage_name}/"
+        return cls.systems_root_uri + f"{system_id}/{storage_name}"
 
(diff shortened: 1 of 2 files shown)


prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -287,21 +287,28 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
...
 
                 # list of storage controllers for that storage id
-                storage_controllers_list: List[Dict] = self.redfish_obj.get(curr_storage_uri).dict[
-                    "StorageControllers"
-                ]
-                storage_controller_count[system_id] += len(storage_controllers_list)
+                storage_controllers = self.redfish_obj.get(curr_storage_uri).dict
+                if "StorageControllers" in storage_controllers:
+                    storage_controllers_list: List[Dict] = storage_controllers.get("StorageControllers")
+                if "Controllers" in storage_controllers:
```
rgildein, line 297: With this approach we can end up with `storage_controllers_list` undefined. We should raise some exception in with else.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -287,21 +287,28 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
...
+                    storage_controllers_list: List[Dict] = []
+                    controller_list = storage_controllers.get("Controllers")
+                    for controller in controller_list.values():
+                        all_controller_data = self.redfish_obj.get(controller).dict
+                        for controller in all_controller_data["Members"]:
+                            storage_controllers_list.append(self.redfish_obj.get(controller["@odata.id"]).dict)
```
rgildein, line 303: I do not think that it's good to have such nested for loop. Can you create simple function instead of doing it directly here.

rgildein (changes requested):
You haven't added any unit tests, please add them. Thanks

chanchiwai-ray:
Can you provide the iLO version and the vendor (Inspur)'s official documentation for Redfish ?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -221,7 +221,29 @@ def _storage_root_uri(cls, system_id: str, storage_name: str) -> str:
...
+        """Return storage controller list.
+
+        Returns:
+            storage_controller_list: list of storage controllers for 
+            requested storage id.
+
+        """
+        if "StorageControllers" in storage_controllers:
+            storage_controllers_list: List[Dict] = storage_controllers.get("StorageControllers")
+        elif "Controllers" in storage_controllers:
```
dashmage, line 236: docs: Could we leave a comment here mentioning about the ILO version/ vendor that requires the code to handle the "Controllers" key?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -221,7 +221,29 @@ def _storage_root_uri(cls, system_id: str, storage_name: str) -> str:
           system_id: "S1"
           storage_name" "Storage"
         """
-        return cls.systems_root_uri + f"{system_id}/{storage_name}/"
+        return cls.systems_root_uri + f"{system_id}/{storage_name}"
+
+    def _find_storage_controller_data(self, storage_controllers: Dict) -> List[Dict]:
```
dashmage, line 226: Naming the parameter here `storage_controllers` is a bit misleading. When called like 

```
self._find_storage_controller_data(
                    self.redfish_obj.get(curr_storage_uri).dict
                )
```

it would contain all the data from that particular storage id since `curr_storage_uri` would be `/redfish/v1/Systems/s1/Storage/Stor1` for example. Maybe something along the lines of `storage_data` could sound more appropriate?

The if condition would also make more sense:
```
if "StorageControllers" in storage_data:
        storage_controllers_list: List[Dict] = storage_data.get("StorageControllers")
```

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -221,7 +221,29 @@ def _storage_root_uri(cls, system_id: str, storage_name: str) -> str:
...
+            requested storage id.
+
+        """
+        if "StorageControllers" in storage_controllers:
+            storage_controllers_list: List[Dict] = storage_controllers.get("StorageControllers")
+        elif "Controllers" in storage_controllers:
+            storage_controllers_list: List[Dict] = []
+            controller_list = storage_controllers.get("Controllers")
+            for controller in controller_list.values():
+                all_controller_data = self.redfish_obj.get(controller).dict
```
dashmage, line 240: docs: It would also be nice to have a comment here to understand what we're expecting the `controller_list` to look like, to better understand why we need another nested for loop querying for the `Members` key.

Eg: Maybe we could mention that `Controllers` only returns a list of links to the storage controllers when fetched from `Members`. And to get individual storage controller data, we need to query via separate GET requests to the URI pointed by `@odata.id`?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -287,21 +309,22 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
...
                 storage_controller_count[system_id] += len(storage_controllers_list)
 
                 # picking out the required data from each storage controller in the list
                 for data in storage_controllers_list:
+                    print(data)
                     storage_controller_data_in_curr_system.append(
                         {
                             "storage_id": storage_id,
-                            "controller_id": data["MemberId"],
+                            "controller_id": data["MemberId"] if data.get("MemberId", None) else data.get("Id"),
```
dashmage, line 327: docs: A comment here as well mentioning that `Id` is what we expect if the key being used is `Controllers` in the storage data would be helpful

tests/unit/test_redfish.py
```diff
@@ -485,7 +485,7 @@ def test__storage_root_uri(self):
     @patch(
         "prometheus_hardware_exporter.collectors.redfish.redfish_utilities.systems.get_system_ids"
     )
-    def test_get_storage_controller_data_success(
+    def test_get_storage_controller_data_success_with_storagecontrollers(
```
dashmage, line 488: nit: Maybe `test_get_storage_controller_data_success` and `test_get_storage_controller_data_success_with_non_standard_api` could be more descriptive test names?

dashmage (changes requested):
Looks quite nice overall, just a few comments. Thanks for submitting this PR!

chanchiwai-ray:
Thanks @sdemircan for the work, this has been superseded by #71. Feel free to report further issues if you find any !

PR #60, opened 2024-02-06
Fix charm name in readme

sudeephb:


agileshaw (approved):
Non-blocking: it would be nice if we can also include the Charmhub link in the readme for charm

sudeephb:
> Non-blocking: it would be nice if we can also include the Charmhub link in the readme for charm
Thanks, I've added it.

rgildein (approved):

PR #64, opened 2024-03-06
Refactor redfish login and logout methods. 

chanchiwai-ray:
The redfish context manager could potentially failed during `__enter__` which might causes the `__exit__` function never called. We should ensure the redfish client is logged out in any exception. This can be achieved by adding a finally block to the try-except block.

Note that even if the `login()` crashes, the server might still recognize the login as successful, and a force logout might not work since we did not get the response from the initial `login()`. For example, if a login is timed out, and the exporter service raise and catch the exception, but the response from the `login()` will not set the `session_key` or `session_location` in the response object (since we never get back a response), so a force `logout()` will not be able to perform a `DELETE` (using the `session_location`) on the server to close the connection. This is an inherent issue on the redfish library that probably cannot be easily fixed

Closes: https://github.com/canonical/hardware-observer-operator/issues/136


prometheus_hardware_exporter/collector.py
```diff
@@ -995,42 +995,51 @@ def fetch(self) -> List[Payload]:
...
+            payloads.append(Payload(name="redfish_call_success", value=0.0))
+            return payloads
+        except Exception as err:  # pylint: disable=W0718
+            logger.exception("Exception occurred while using redfish object: %s", err)
             payloads.append(Payload(name="redfish_call_success", value=0.0))
             return payloads
+        finally:
+            # Guarding against `RedfishHelper` crashes, and `redfish_helper` remains `None`
+            if redfish_helper:
+                redfish_helper.logout()
```
rgildein, line 1042: If the context manager was not logout, I believe that we should fix it there and not here. Why defining as context manager, which is not working correctly.

rgildein: The `RedfishHelper` is our object, I hope so that I'm not wrong.

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -47,10 +48,17 @@ def _discover(host: str) -> bool:
...
+                result = False
             else:
                 # redfish available, login succeeded
                 result = True
-                redfish_obj.logout()
+            finally:
+                # make sure we logout in the end
+                if redfish_obj:
+                    redfish_obj.logout()
+                    logger.debug("(discover) Logged out from redfish service ...")
```
rgildein, line 61: Checking briefly the [redfish source code](https://github.com/DMTF/python-redfish-library/blob/main/src/redfish/rest/v1.py#L502), the client is also context manager, so why using it?

rgildein (changes requested):

prometheus_hardware_exporter/collector.py
chanchiwai-ray: Can you check the PR description again, and see if you can understand the explanation I provided. Short answer is if `__enter__`  crashes, `__exit__` will not run. Hence, we are fixing it here

chanchiwai-ray: Example:
```python3
class Foo:

    def __enter__(self):
        print("enter")
        raise KeyError

    def __exit__(self, *args, **kwargs):
        print("exit")

try:
    with Foo() as f:
        print("do something")
except ValueError:
    print("catch ValueErrror")
except Exception:
    print("catch Exception")
```

prometheus_hardware_exporter/collectors/redfish.py
chanchiwai-ray: It's a bit convoluted now, the `RedfishHelper` wraps `redfish_client`, so we won't directly use it's context manager, plus it will suffer the same issue as described in the description

prometheus_hardware_exporter/collector.py
rgildein: Now I see, but still is strange behavior. Why should move that try except to the __enter__ and not outside. To me, we are not defining context manager properly.

Why we are defining redfish_obj in __enter__ and not in __init__? Like this if init failed you do not need to run logout (and try except make sense) and if login failed you should not do any logout again or?
```python
 def __init__(self, config: Config) -> None:
        """Initialize values for class."""
        self.host = config.redfish_host
        self.username = config.redfish_username
        self.password = config.redfish_password
        self.redfish_obj = redfish_client(
            base_url=self.host,
            username=self.username,
            password=self.password,
            timeout=self.timeout,
            max_retry=self.max_retry,
        )

    def __enter__(self) -> Self:
        """Login to redfish while entering context manager."""
        self.redfish_obj.login(auth="session")
        return self
```

chanchiwai-ray: Yes, that's one of the option I discussed with @dashmage : adding `try-except` in the `__enter__`, but that essentially just making a nested `try-except` block. So I guess both this implementation and the one you suggested has pros and cons (in terms of the coding style, no right or wrong )

chanchiwai-ray:
Note that I edit my PR description indicating a potential loop hole that we can't really fix. The workaround is to increase the redfish timeout so that the case I mentioned can be significantly reduced.

rgildein (approved):
Do not agree with defining the context manager, but not using it, since it can fail. But I do not see this blocker issue for now.

chanchiwai-ray:
> Do not agree with defining the context manager, but not using it, since it can fail. But I do not see this blocker issue for now.

We shouldn't simply remove it because we don't use it, since it will cause a backward compatibility issue (just in case other users are using this repo as a library).

rgildein:
> > Do not agree with defining the context manager, but not using it, since it can fail. But I do not see this blocker issue for now.
> 
> 
> 
> We shouldn't simply remove it because we don't use it, since it will cause a backward compatibility issue (just in case other users are using this repo as a library). 

I never suggested that, but if anyone is using as lib, they will hit same issue and the fix will be same as you are doing "not using context manager".

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -67,29 +75,32 @@ def __init__(self, config: Config) -> None:
...
         self.redfish_obj = redfish_client(
             base_url=self.host,
             username=self.username,
             password=self.password,
             timeout=self.timeout,
             max_retry=self.max_retry,
         )
-        self.redfish_obj.login(auth="session")
+
+    def __enter__(self) -> Self:
```
gabrielcocenza, line 86: If I understood correctly you are wrapping redfish_client in this class. As I said on this [comment](https://github.com/canonical/hardware-observer-operator/pull/182#discussion_r1514641149) on `hardware-observer`, I consider this as a bug on the upstream. However, if you are wrapping it, this might be a good opportunity to implement what is necessary to run the disconnect using `__exit__` and pretty much is not necessary to change the `collector.py` file.

dashmage:
After reading through the comments and thinking about this,
* Currently -- we have a context manager to manage the login/logout for redfish as part of the `RedfishHelper` class. We also need to have a try/except block around the context manager itself to catch exceptions while running the various helper methods for redfish (`get_chassis_data`, `get_storage_drive_data` etc.) inside it.
* The main issue is when an exception occurs while executing `redfish_obj.login` which will cause `__enter__` to raise it and `__exit__` isn't called. Which is supposed to be fine...but somehow there's still a login session being created so not logging out is a problem. As mentioned in the existing comments already, there does seem to be an upstream bug lurking here.
* All we'd need to do is force a logout -- this can happen with a try/except in `__enter__` similar to what @gabrielcocenza proposed. Or we could just directly call the login/logout methods in `collector.py` so there's just one set of try/except/finally statements and we can ignore the context manager -- which is Ray's implementation in the PR now. I'm personally in favor of the second option.
* I don't quite understand how if the `redfish_obj.login` doesn't work and the `session_{key,location}` aren't set, the server still registers a login. If `redfish_obj.logout` itself doesn't work for some reason to clean up the session, then the fix would certainly have to come from the upstream redfish library. We could potentially check for existing sessions and perform cleanup manually but then it would run the risk of removing legitimate sessions by actual users since we wouldn't have any way of identifying the sessions created by the charm.

Please correct me if I'm wrong in any of my assumptions in my summary.

chanchiwai-ray:
> I don't quite understand how if the redfish_obj.login doesn't work and the session_{key,location} aren't set, the server still registers a login. If redfish_obj.logout itself doesn't work for some reason to clean up the session, then the fix would certainly have to come from the upstream redfish library. We could potentially check for existing sessions and perform cleanup manually but then it would run the risk of removing legitimate sessions by actual users since we wouldn't have any way of identifying the sessions created by the charm.

For example if you set a login timeout to 1s, and you think your server can handles the login in 1s, that's fine. However, if the server take too long to respond (say 5s), then `redfish_client` raise time out exception, and can't handle the response which sets the `self.__session_key`. In that case, even if we do `redfish_client.logout` [1] it will do nothing.

[1] https://github.com/DMTF/python-redfish-library/blob/main/src/redfish/rest/v1.py#L1027

chanchiwai-ray:
Rebased onto [ba509b6](https://github.com/canonical/prometheus-hardware-exporter/commit/ba509b6124fe6b1d4947303788a391c14f57b063)

prometheus_hardware_exporter/collectors/redfish.py
chanchiwai-ray: Yes, I agree, that's another choice we already mentioned and can make, but decided not to introduce `try-except` in `__enter__`, which effectively doing a nested `try-except` block.

```python3
try:
    with RedfishHelper(...) as redfish_helper: <-- another try-except inside or even fix the upstream
....
```

This solution will not introduce less code that the current implementation: adding `try-except` in `__enter__` vs adding `finally` in current `try-except`

(if you exclude my changes to unit test (which I probably should not change, but I did) that remove the context manager)

dashmage (approved):
LGTM overall, thanks :)

prometheus_hardware_exporter/collector.py
```diff
@@ -995,41 +995,51 @@ def fetch(self) -> List[Payload]:
...
+            smart_storage_health_data: Dict[str, Any] = (
+                redfish_helper.get_smart_storage_health_data()
+            )
         except (
             ConnectionError,
             InvalidCredentialsError,
             RetriesExhaustedError,
             SessionCreationError,
         ) as err:
-            logger.exception("Exception occurred while getting redfish object: %s", err)
```
gabrielcocenza, line 1030: nit: I don't see any difference from the expected exceptions to the generic one. Can't you simply add `Exception` to the multiple exceptions?

gabrielcocenza (approved):
LGTM. Just a small comment that it's not relevant.

prometheus_hardware_exporter/collector.py
chanchiwai-ray: Done

Pjack (approved):

PR #70, opened 2024-04-22
Recreate sdr cache if invalid

### LINKED ISSUE #40: Flush cache while running ipmi commands
Sometimes the outdated cache may cause error and cause the service to stop. We need to flush the cache when the service starts.
The cache may be helpful to make the collector faster, so we can discuss further on where to and where not to flush the cache.

IPMI provides a `-f` command to flush cache while running commands.

sudeephb:
This recreates the ipmi sdr cache if it's out of date or invalid.
If the sdr cache is working, nothing changes.

Closes #40

### DIFF
--- a/prometheus_hardware_exporter/collectors/ipmi_sel.py
+++ b/prometheus_hardware_exporter/collectors/ipmi_sel.py
@@ -23,6 +23,9 @@ class IpmiSel(Command):
         Returns:
             sel_entries: a list of dictionaries containing sel_sentries, or []
         """
+        # --sdr-cache-recreate is required to automatically recreate the SDR cache in case it is
+        # out of date or invalid. Without this, the service will stop getting ipmi-sel data if the
+        # cache is out of date.
         result = self(
             "--sdr-cache-recreate --output-event-state --interpret-oem-data --entity-sensor-names"
         )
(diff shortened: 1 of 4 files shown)


aieri (approved):

prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -23,7 +23,9 @@ def get_sel_entries(self, time_range: int) -> Optional[List[Dict[str, str]]]:
         Returns:
             sel_entries: a list of dictionaries containing sel_sentries, or []
         """
-        result = self("--output-event-state --interpret-oem-data --entity-sensor-names")
+        result = self(
+            "--sdr-cache-recreate --output-event-state --interpret-oem-data --entity-sensor-names"
```
jneo8, line 27: **nitpick (blocking):** 

Can you add some comments the mention why those options is required and also add unit-test to it?

dashmage (approved):
LGTM! Similar to what James mentioned, if you could add a comment explaining the options/ provide link to some docs, that would be super helpful for anyone going through this :smile:

prometheus_hardware_exporter/collectors/ipmi_sel.py
sudeephb: Thanks, I have added comments on why the flag is needed. 
For the unit-test, however, [we already have tests that mock __call__](https://github.com/canonical/prometheus-hardware-exporter/blob/6c55a2d4665480204060bfcd09512e702f0e7c45/tests/unit/test_ipmi_sel.py#L49), which mock this call to `ipmi-sel` and `ipmimonitoring` in this case, so the flags don't really matter.

jneo8 (approved):

PR #74, opened 2024-04-29
Fix regression from #71.

chanchiwai-ray:
Set the "Health" key to a string of "NA" when it's falsy (e.g. `None`, "").

Note: 

```
d.get("key") or "NA"  # return NA if d = {"key": <falsy-value>}
d.get("key", "NA")     # return None if d = {"key": <falsy-value>}
```

If "Health" is `None`, it can cause the following error:

```
AttributeError: ("'NoneType' object has no attribute 'replace'", Metric(redfish_storage_controller, Storage Controller information obtained from redfish., info, , [Sample(name='redfish_storage_controller_info', labels={'system_id': 'System.Embedded.1', 'storage_id': 'AHCI.Embedded.1-1', 'controller_id': '0', 'health': None, 'state': 'Enabled'}, value=1, timestamp=None, exemplar=None)]))
```

### DIFF
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -331,8 +331,10 @@ class RedfishHelper:
                 for data in storage_controllers_list:
                     # "Id" is what we expect if the key being used is Controllers
                     controller_id = data.get("MemberId") or data.get("Id")
-                    state = data.get("Status", {}).get("State")  # when is falsy set to None
-                    health = data.get("Status", {}).get("Health") or "NA"  # when is falsy set NA
+                    state = data.get("Status", {}).get("State")  # when it's falsy set to None
+                    health = (
+                        data.get("Status", {}).get("Health") or "NA"
+                    )  # when it's falsy set to NA
                     if not controller_id or not state:  # health is not required
                         logger.warning(
                             "No relevant data found in storage controller data: %s", data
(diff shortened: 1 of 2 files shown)


sudeephb:
Can you add a unit test covering the scenario where "Health" is not present?

jneo8 (approved):

sudeephb (approved):
Thanks, LGTM

sudeephb (approved):

chanchiwai-ray:
Sorry, @sudeephb @jneo8 I added one more test case, can you check again ?

prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -330,10 +330,10 @@ def get_storage_controller_data(self) -> Tuple[Dict[str, int], Dict[str, List]]:
                 # picking out the required data from each storage controller in the list
                 for data in storage_controllers_list:
                     # "Id" is what we expect if the key being used is Controllers
-                    controller_id = data.get("MemberId", "") or data.get("Id", "")
-                    state = data.get("Status", {}).get("State", None)
-                    health = data.get("Status", {}).get("Health", "NA")
-                    if controller_id == "" or not state:  # health is not required
+                    controller_id = data.get("MemberId") or data.get("Id")
+                    state = data.get("Status", {}).get("State")  # when is falsy set to None
+                    health = data.get("Status", {}).get("Health") or "NA"  # when is falsy set NA
```
dashmage, line 335:
```suggestion
                    state = data.get("Status", {}).get("State")  # when it's falsy set to None
                    health = data.get("Status", {}).get("Health") or "NA"  # when it's falsy set NA
```

dashmage (approved):
Just one tiny nit on the comment, LGTM!

Issue #76, opened 2024-06-25
Invalid session resource errors cause CollectorFailed alerts


peterctl:
Using the [`hardware-observer` charm](https://github.com/canonical/hardware-observer-operator), the [`CollectorFailed`](https://github.com/canonical/hardware-observer-operator/blob/main/src/prometheus_alert_rules/general.yaml) alerts fire multiple times a day, generating a lot of noise for the alerts on-call. The alerts seem to match with these errors showing up in the systemd service logs:
```
ubuntu@hostname:~$ journalctl -u hardware-exporter.service | grep ERROR | tail -n 10
Jun 25 16:54:35 hostname python3[$PID]: 2024-06-25 16:54:35 ERROR Invalid session resource: /redfish/v1/SessionService/Sessions/102390, return code: 401
```

aieri:
we should look into having the collector try to refresh the session a few times before logging the error - i.e. those log lines should probably not make it to syslog at all

peterctl:
Here are some more logs around the error message for extra context.
```bash
# journalctl -u hardware-exporter.service | grep -B30 ERROR
Jul 16 13:59:30 hostname python3[3681]: 2024-07-16 13:59:30 INFO Login returned code 201: {"@Message.ExtendedInfo":[{"Message":"The resource has been created successfully.","MessageArgs":[],"MessageArgs@odata.count":0,"MessageId":"Base.1.12.Created","RelatedProperties":[],"RelatedProperties@odata.count":0,"Resolution":"None.","Severity":"OK"},{"Message":"A new resource is successfully created.","MessageArgs":[],"MessageArgs@odata.count":0,"MessageId":"IDRAC.2.8.SYS414","RelatedProperties":[],"RelatedProperties@odata.count":0,"Resolution":"No response action is required.","Severity":"Informational"}],"@odata.context":"/redfish/v1/$metadata#Session.Session","@odata.id":"/redfish/v1/SessionService/Sessions/132562","@odata.type":"#Session.v1_6_0.Session","ClientOriginIPAddress":"IP_FOR_BOND_M","CreatedTime":"2024-07-16T13:59:10-05:00","Description":"User Session","Id":"132562","Name":"User Session","Password":null,"SessionType":"Redfish","UserName":"root"}
Jul 16 13:59:30 hostname python3[3681]: 2024-07-16 13:59:30 INFO Getting redfish sensor info...
[...]

PR #78, opened 2024-07-17
Ignore if redfish fails to logout

gabrielcocenza:
Redfish sometimes fails to logout after scraping the data and this end up generating the metric `redfish_collector_failed` that trigger prometheus alerts that can be noise for production clouds.

This can be a BMC problem that might not be properly closing sessions.

For more info see:
https://github.com/DMTF/python-redfish-library/issues/160

Closes: #76

### DIFF
--- a/prometheus_hardware_exporter/collectors/redfish.py
+++ b/prometheus_hardware_exporter/collectors/redfish.py
@@ -7,6 +7,7 @@ import redfish_utilities
 from cachetools.func import ttl_cache
 from redfish import redfish_client
 from redfish.rest.v1 import (
+    BadRequestError,
     InvalidCredentialsError,
     RetriesExhaustedError,
     SessionCreationError,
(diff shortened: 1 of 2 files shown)


prometheus_hardware_exporter/collectors/redfish.py
```diff
@@ -92,8 +93,16 @@ def login(self) -> None:
...
-        logger.debug("(service) Logged out from redfish service ...")
+        # redfish sometimes fails to logout and this end up giving false alerts
+        # See https://github.com/DMTF/python-redfish-library/issues/160 and
+        # https://github.com/canonical/prometheus-hardware-exporter/issues/76
+        try:
+            self.redfish_obj.logout()
+            logger.debug("(service) Logged out from redfish service ...")
+        except BadRequestError as err:
+            logger.error("Failed to logout redfish: %s", str(err))
+            if "code: 401" not in str(err):
```
Pjack, line 104: We might consider ignoring all bad request errors during logout. It's a weird situation and nothing we can do. I don't want to trigger exception and impact the collector service.

jneo8: +1, please consider the situation that redfish is a API endpoint and sometimes you wouldn't able to connect to it due to multiple possible reasons.

Warning message should be enough in our case?

jneo8 (changes requested):

chanchiwai-ray:
I don't know what's the best choice, if redfish can't logout properly, in the long run it will still cause issue like this https://github.com/canonical/hardware-observer-operator/issues/136

Pjack (approved):

jneo8 (approved):

PR #79, opened 2024-07-18
Simplify the fetch exception on redfish

gabrielcocenza:
When any exception happens during redfish fetch the metric redfish_call_success is set to 0 that will trigger Prometheus alerts.


chanchiwai-ray (approved):
LGTM

aieri (approved):

PR #82, opened 2024-09-13
Remove unused/old files

samuelallan72:
- we don't use Sonar any more
- Makefile and rename.sh are also not really used; use tox.ini or `snapcraft` directly.
- update check.yaml to follow our latest conventions 
- we don't really use pre-commit any more, or the tox dev-environment target


.github/workflows/check.yaml
```diff
@@ -15,22 +15,82 @@ concurrency:
...
+  lint:
+    name: Lint
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.10"]
-    with:
-      python-version: ${{ matrix.python-version }}
-      tox-version: "<4"
+    runs-on: ubuntu-latest
```
gabrielcocenza, line 24: maybe this should be 22.04 to follow the other repos?

samuelallan72: thanks, updated :)

chanchiwai-ray (approved):

gabrielcocenza (approved):

Issue #62, opened 2024-02-14
ipmi_sel collector uses an arbitrary and non-configurable 30s timeout which results in false positive alerts


przemeklal:
Hi,

On systems with many entries in SEL (500+ based on my observations), it may take more than 30s to collect ipmi-sel output:

```
Feb 14 13:00:29 maas1 python3[2177132]: 2024-02-14 13:00:29 ERROR Command 'ipmi-sel --output-event-state --interpret-oem-data --entity-sensor-names' timed out after 30 seconds
```

Running the same command manually works fine but indeed it takes more than 30 seconds:
```
time ipmi-sel --output-event-state --interpret-oem-data --entity-sensor-names
...
real    0m32.664s
user    0m0.140s
sys    0m0.176s
```

This threshold doesn't seem to be configurable. It is set too low by default and causes many false positive alerts.

One of the possible workarounds is to clear SEL which reduces the time needed to execute `ipmi-sel --output-event-state --interpret-oem-data --entity-sensor-names` although in my opinion the collector shouldn't fail just because there are 500+ entries in SEL.

chanchiwai-ray:
Thanks for the report, the timeout is perhaps set by the [subprocess](https://github.com/canonical/prometheus-hardware-exporter/blob/main/prometheus_hardware_exporter/utils.py#L65) called which is hardcoded. To you suggestion, I am not sure if increasing the timeout or making the timeout configurable by end user is a viable solution or not, since the Prometheus will scrape periodically, and if the timeout is longer than the scrape period, it might cause more trouble than being useful... 

We will discuss internally about this on how to address it.

przemeklal:
Ok, just please be aware that this issue is widespread, almost every cloud where hardware-exporter is deployed suffers from this issue.

jneo8:
Hi @przemeklal 

I would like to ask some of the behaviors we want to achieve here.

1. If the metrics is not real time, can it be acceptable?
2. If it spend too much time on querying(more than prometheus timeout), can missing metrics be acceptable?

przemeklal:
1. It depends on how it's implemented. If you plan to add cache then you need to make sure that it's not too old. This means more alert rules, another arbitrary timeout elsewhere, etc. Also, aren't there some alert rules that look back only a certain amount of time? Wouldn't we risk not detecting them?
2. There are rules that alert on missing metrics now, so I think the answer is no. We'd also risk not getting any alerts when the exporter/collector dies.

[...]

PR #91, opened 2025-03-04
Introduce a collect thread to IpmiSel

Deezzir:
Closes: #62


prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -15,6 +18,46 @@ class IpmiSel(Command):
...
+        """Initialize the IpmiSel class."""
+        super().__init__(config)
+        self.config.collect_timeout = None
+        self._cache: Optional[List[Dict[str, str]]] = []
+        self._lock = threading.Lock()
+        self._update_thread: Optional[threading.Thread] = None
+        self._thread_started = False
+
+    def _update_cache(self) -> None:
+        """Background thread function to update SEL cache periodically."""
```
samuelallan72, line 31: Won't this only run once? There's no loop in the method.

prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -15,6 +18,46 @@ class IpmiSel(Command):
...
+            raw_sel_data = result.data.strip().split("\n")
+            sel_data_fields = ["ID", "Date", "Time", "Name", "Type", "State", "Event"]
+            sel_entries = []
+            for sel_item in raw_sel_data[1:]:
+                sel_item_values = [entry.strip() for entry in sel_item.split("|")]
+                sel_item_dict = dict(zip(sel_data_fields, sel_item_values))
+                sel_entries.append(sel_item_dict)
+
+            with self._lock:
+                self._cache = sel_entries
```
samuelallan72, line 48: The cache must include some kind of timestamp for when the cache was populated, and then when returning the cached results, we must check if the timestamp is within some time-to-live. Otherwise if something goes wrong, it can be serving stale data forever.

prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -15,6 +18,46 @@ class IpmiSel(Command):
...
+        else:
+            logger.error("Failed to fetch SEL entries: %s", result.error)
+            self._cache = None
+
+        time.sleep(self.config.ipmi_sel_collect_interval)
```
samuelallan72, line 53: This doesn't take into account how long the collector command tool to run - this may be significant if it can take > 30 seconds. Is it important to have a more consistent interval?

prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -15,6 +18,46 @@ class IpmiSel(Command):
     prefix = ""
     command = "ipmi-sel"
 
+    def __init__(self, config: Config) -> None:
```
samuelallan72, line 21: How long-lived is this class in the application? Will be it be recreated at any point? Curious because this may change how the thread is implemented.

prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -15,6 +18,46 @@ class IpmiSel(Command):
...
+            logger.error("Failed to fetch SEL entries: %s", result.error)
+            self._cache = None
+
+        time.sleep(self.config.ipmi_sel_collect_interval)
+
+    def _start_thread(self) -> None:
+        """Start the background thread to update SEL cache."""
+        self._update_thread = threading.Thread(target=self._update_cache, daemon=True)
+        self._update_thread.start()
+        self._thread_started = True
```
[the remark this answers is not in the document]
samuelallan72: (although if the thread is started immediately on class init, then we don't need to check this)

chanchiwai-ray (commented):
As @jneo8 mentioned, this change is better to apply to the collector level. You can easily create new `ThreadedCollector(Collector)` (or whatever name that describes the purpose) in https://github.com/canonical/prometheus-hardware-exporter/blob/45c01ab449dfa8384cd37398f7dd1b04398599a0/prometheus_hardware_exporter/core.py#L40. to add this feature

chanchiwai-ray (dismissed):
LGTM

prometheus_hardware_exporter/collector.py
```diff
@@ -591,14 +620,18 @@ def specifications(self) -> List[Specification]:
 
     def fetch(self) -> List[Payload]:
         """Load ipmi sel entries."""
-        sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
+        if self.is_cache_expired():
```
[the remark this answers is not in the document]
chanchiwai-ray: not a blocker, feel free to resolve

prometheus_hardware_exporter/collector.py
```diff
@@ -564,14 +567,40 @@ def _get_sensor_value_from_reading(self, reading: str) -> float:
...
+            start_time = datetime.datetime.now()
+
+            sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
+            with self._lock:
+                self._cache = sel_entries
+            self._cache_timestamp = datetime.datetime.now().timestamp()
+
+            elapsed_time = (datetime.datetime.now() - start_time).total_seconds()
+            sleep_time = max(0, self.config.ipmi_sel_interval - elapsed_time)
+            time.sleep(sleep_time)
```
jneo8, line 595: **suggestion (non-blocking):** 

Since *Command* object already provide some error handling, I think the code is good. But it will be more nicer if you can add some logging message here when every time cache is been updated or failed to fetch.

jneo8 (approved):
Overall LGTM.

(One not blocking improvement suggestion)

prometheus_hardware_exporter/collector.py
Deezzir: I was just thinking about that, addressing

prometheus_hardware_exporter/collectors/ipmi_sel.py
Deezzir: Addressed

prometheus_hardware_exporter/collectors/ipmi_sel.py
Deezzir: Addressed

prometheus_hardware_exporter/collectors/ipmi_sel.py
Deezzir: The class is only created once on the startup

prometheus_hardware_exporter/collectors/ipmi_sel.py
```diff
@@ -23,32 +66,20 @@ def get_sel_entries(self, time_range: int) -> Optional[List[Dict[str, str]]]:
...
-        # out of date or invalid. Without this, the service will stop getting ipmi-sel data if the
-        # cache is out of date.
-        result = self(
-            "--sdr-cache-recreate --output-event-state --interpret-oem-data --entity-sensor-names"
-        )
-        if result.error:
-            logger.error(result.error)
-            return None
+        if not self._thread_started:
+            self._start_thread()
```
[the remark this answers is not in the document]
Deezzir: I've decided to move the thread startup to the `__init__` method, thanks!

chanchiwai-ray (dismissed):

chanchiwai-ray (approved):

jneo8 (dismissed):

prometheus_hardware_exporter/collector.py
```diff
@@ -564,14 +567,43 @@ def _get_sensor_value_from_reading(self, reading: str) -> float:
...
+        logger.info("Starting IPMI SEL cache update thread.")
+
+        while True:
+            start_time = datetime.datetime.now()
+
+            sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
+            with self._lock:
+                self._cache = sel_entries
+            self._cache_timestamp = datetime.datetime.now().timestamp()
+            logger.info("Updated IPMI SEL cache.")
```
gabrielcocenza, line 594: I have a concern that if the `get_sel_entries` fails for some reason the while True loop will stop executing because the thread dies andnNo further cache updates will occur, as the background thread is no longer running.

I think it's safer to add a try/except and log the error. Something like:

```python
try:
                sel_entries = self.ipmi_sel.get_sel_entries(self.config.ipmi_sel_interval)
                with self._lock:
                    self._cache = sel_entries
                self._cache_timestamp = datetime.datetime.now().timestamp()
                logger.info("Updated IPMI SEL cache.")
            except Exception as e:
                logger.error(f"Failed to update IPMI SEL cache: {str(e)}")
```

gabrielcocenza (approved):

jneo8 (approved):

chanchiwai-ray (approved):

PR #119, opened 2026-01-29
feat: ipmitool power supply over lan

gabrielcocenza:
Not having ipmitool power supply over lan can timeout the collector.

Unify the cli arguments to be usable for both redfish and ipmi


H-M-Quang-Ngo (approved):
LGTM!

Pjack (approved):

jneo8 (approved):

